PR #3, opened 2016-02-01
Implemented loop parameter for samples

bgold-cosmos:
I use `loop` a lot, so I implemented it in SuperDirt.  My apologies if I didn't do it in the best way, I'm still fairly new to sclang...


bgold-cosmos:
Thanks for the catch, should be good now

telephon:
thank you!

PR #14, opened 2016-05-28
Default synth library

bgold-cosmos:
5 tutorial examples, 5 new tonal synths (mostly hooks into physical
modeling UGens), and 5 analog-style drum synths. Other than the tutorial examples,
all begin with "super" in the name.  They are:

```
supermandolin
superpiano
supergong
superhex
supersiren
superkick
supersnare
superhat
superclap
super808
```

### DIFF
--- a/synths/default-synths.scd
+++ b/synths/default-synths.scd
@@ -124,7 +124,7 @@ SynthDef(\tutorial4, {|out, begin=0, end=1, pan, speed=1, accelerate, n |
 }).add;
 
 // finally, as an example of how to add your own parameters, let's say we want precise control over frequency
-// we'll make a new "f" parameter instead of "n" (don't use "freq", it's already used internally in SuperDirt)
+// we'll make a new "f" parameter instead of "n"
 // to do this, we first need to this in Tidal: let (f, f_p) = pF "f" (Just 440)
 // then "f" is usable as an argument to the synthdef, and in Tidal you can try
 // d1 $ s "tutorial5/2" # f "[550,555]" # speed "1"
--- a/synths/default-synths.scd
+++ b/synths/default-synths.scd
@@ -107,14 +107,14 @@ SynthDef(\tutorial2, {|out, sustain, pan |
 // in Tidal we'll need to say something like `sound "tutorial3:69"` to hear a reasonably high pitch
 SynthDef(\tutorial3, {|out, sustain, pan, accelerate, n |
 	var env = EnvGen.ar(Env.linen(0.01, 0.98, 0.01, 1, -3), timeScale:sustain, doneAction:2);
-	var sound = SinOsc.ar(n.midicps * Line.kr(1,1+accelerate));
+	var sound = SinOsc.ar(n.midicps * Line.kr(1,1+accelerate, sustain));
 	OffsetOut.ar(out, DirtPan.ar(sound, ~dirt.numChannels, pan, env));
 }).add;
 
 // we can also make the envelope a more interesting percussive shape
 SynthDef(\tutorial4, {|out, sustain, pan, accelerate, n |
 	var env = EnvGen.ar(Env.perc(0.001, 0.999, 1, -4), timeScale:sustain, doneAction:2);
-	var sound = SinOsc.ar(n.midicps * Line.kr(1,1+accelerate));
+	var sound = SinOsc.ar(n.midicps * Line.kr(1,1+accelerate, sustain));
 	OffsetOut.ar(out, DirtPan.ar(sound, ~dirt.numChannels, pan, env));
 }).add;
 
@@ -125,7 +125,7 @@ SynthDef(\tutorial4, {|out, sustain, pan, accelerate, n |
 // d1 $ s "tutorial5/2" # f "[550,555]"
 SynthDef(\tutorial5, {|out, sustain, pan, accelerate, f |
 	var env = EnvGen.ar(Env.perc(0.001, 0.999, 1, -4), timeScale:sustain, doneAction:2);
-	var sound = SinOsc.ar(f * Line.kr(1,1+accelerate));
+	var sound = SinOsc.ar(f * Line.kr(1,1+accelerate, sustain));
 	OffsetOut.ar(out, DirtPan.ar(sound, ~dirt.numChannels, pan, env));
 }).add;
 
@@ -135,7 +135,7 @@ SynthDef(\tutorial5, {|out, sustain, pan, accelerate, f |
 SynthDef(\supermandolin, {|out, sustain, pan, accelerate, n, detune=0.2 |
 	var env = EnvGen.ar(Env.linen(0.002, 0.996, 0.002, 1,-3), timeScale:sustain, doneAction:2);
 	var sound = Decay.ar(Impulse.ar(0,0,0.1), 0.1*n/69) * WhiteNoise.ar;
-	var pitch = n.midicps * Line.kr(1, 1+accelerate);
+	var pitch = n.midicps * Line.kr(1, 1+accelerate, sustain);
 	sound = CombL.ar(sound, 0.05, pitch.reciprocal*(1-(detune/100)), sustain)
 	          + CombL.ar(sound, 0.05, pitch.reciprocal*(1+(detune/100)), sustain);
 	OffsetOut.ar(out, DirtPan.ar(sound, ~dirt.numChannels, pan, env))
@@ -173,7 +173,7 @@ SynthDef(\superpiano,{|out, sustain, pan, velocity=1, detune=0.1, muffle=1, ster
 // waveguide mesh, hexagonal drum-like membrane
 SynthDef(\superhex,{|out, speed=1, sustain, pan, n, accelerate |
 	var env = EnvGen.ar(Env.linen(0.02, 0.96, 0.02, 1,-3), timeScale:sustain, doneAction:2);
-	var tension = 0.05*n.midicps/400 * Line.kr(1,accelerate+1);
+	var tension = 0.05*n.midicps/400 * Line.kr(1,accelerate+1, sustain);
 	var loss = 1.0 - (0.01 * speed / n.midicps);
 	var sound = MembraneHexagon.ar(Decay.ar(Impulse.ar(0,0,1), 0.01), tension, loss);
 	OffsetOut.ar(out, DirtPan.ar(sound, ~dirt.numChannels, pan, env))
--- a/synths/default-synths.scd
+++ b/synths/default-synths.scd
@@ -79,7 +79,7 @@ SynthDef(\gabor, { |out, sustain = 0.03, note = 0, amp = 1, pan = 0, width = 0.3
 (
 // list of new SynthDefs, and in parentheses the special parameters they accept
 // ("!" means it's not a standard Tidal param and needs to be defined)
-// most SynthDefs use speed to scale the overall envelope, and many use accelerate to provide a pitch glide
+// most of these SynthDefs use sustain to scale the overall envelope, and many use accelerate to provide a pitch glide
 //
 // tutorial1, tutorial2, tutorial3, tutorial4, tutorial5(f!),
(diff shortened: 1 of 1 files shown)


synths/default-synths.scd
```diff
@@ -124,7 +124,7 @@ SynthDef(\tutorial4, {|out, begin=0, end=1, pan, speed=1, accelerate, n |
 }).add;
 
 // finally, as an example of how to add your own parameters, let's say we want precise control over frequency
-// we'll make a new "f" parameter instead of "n" (don't use "freq", it's already used internally in SuperDirt)
```
telephon: This is fixed here: https://github.com/musikinformatik/SuperDirt/commit/60948262d871026fafc788ffbd7cc911f94d9580

I think we should use `freq` instead of `f`or event `n`(when `n.midicps` is done later), because it is easier to understand and also we can apply harmonic calculations in superdirt better.

What do you think?

telephon:
Excellent, thank you! I've just made a comment about the `freq` parameter, we can use this PR as a way to find a good standard for the future maybe.

synths/default-synths.scd
```diff
@@ -75,3 +75,197 @@ SynthDef(\gabor, { |out, sustain = 0.03, note = 0, amp = 1, pan = 0, width = 0.3
...
+	var sound = SinOsc.ar(440.0);
+	OffsetOut.ar(out, DirtPan.ar(sound, ~dirt.numChannels, 0.5));
+}).add;
+
+// now let's add some better duration control, and support for pan
+// Tidal's synth parameters magically show up as arguments to the SynthDef!
+// overall gain is handled elsewhere by SuperDirt, so we don't need it in the envelope
+SynthDef(\tutorial2, {|out, begin=0, end=1, pan, speed=1 |
+	var dur = (end-begin)/speed;
+	var env = EnvGen.ar(Env.linen(0.01, dur, 0.01, 1,-3), timeScale:dur, doneAction:2);
```
telephon: there is a `sustain` parameter passed to whoever needs it, it is calculated from `(end-begin)/speed`.  Start, end and speed are really only needed if you want to use them to select a range of a parameter (like the phase in a soundfile).

(an example is in the gabor grain)

I don't know what is better in a tutorial.

In SuperCollider by convention (not all code does it that way), the parameter `dur` is the inter-onset-time between grains or notes, and the parameter `sustain` is the total length of the grain or note itself.

synths/default-synths.scd
```diff
@@ -75,3 +75,197 @@ SynthDef(\gabor, { |out, sustain = 0.03, note = 0, amp = 1, pan = 0, width = 0.3
...
+	OffsetOut.ar(out, DirtPan.ar(sound, ~dirt.numChannels, pan, env));
+}).add;
+
+// physical modeling of a vibrating string, using a delay line (CombL) excited by an intial pulse (Impulse)
+// To make it a bit richer, I've combined two slightly detuned delay lines
+// "accelerate" is used for a pitch glide, and "speed" changes the envelope timescale
+SynthDef(\supermandolin, {|out, speed=1, begin=0, end=1, pan, accelerate, n, detune=0.2 |
+	var dur = (end-begin)/speed;
+	var env = EnvGen.ar(Env.linen(0.002, dur, 0.002, 1,-3), timeScale:dur, doneAction:2);
+	var sound = Decay.ar(Impulse.ar(0,0,0.1), 0.1*n/69) * WhiteNoise.ar;
```
telephon, line 137: `n/69*0.1` is intended as it is? "Standard" would be either `(n + 60).midicps` or `1/freq`. But this just depends on how you want it to sound.

synths/default-synths.scd
```diff
@@ -75,3 +75,197 @@ SynthDef(\gabor, { |out, sustain = 0.03, note = 0, amp = 1, pan = 0, width = 0.3
...
+	var dur = (end-begin)/speed;
+	var env = EnvGen.ar(Env.perc(0.001/speed, 1/speed, 1, -4), timeScale:dur, doneAction:2);
+	var sound = SinOsc.ar(f * Line.kr(1,1+accelerate));
+	OffsetOut.ar(out, DirtPan.ar(sound, ~dirt.numChannels, pan, env));
+}).add;
+
+// physical modeling of a vibrating string, using a delay line (CombL) excited by an intial pulse (Impulse)
+// To make it a bit richer, I've combined two slightly detuned delay lines
+// "accelerate" is used for a pitch glide, and "speed" changes the envelope timescale
+SynthDef(\supermandolin, {|out, speed=1, begin=0, end=1, pan, accelerate, n, detune=0.2 |
```
telephon: Normally, `detune` is an offset in Hz. Is there a reason you prefer it this way?

synths/default-synths.scd
```diff
@@ -75,3 +75,197 @@ SynthDef(\gabor, { |out, sustain = 0.03, note = 0, amp = 1, pan = 0, width = 0.3
...
+SynthDef(\supergong,{|out, speed=1, begin=0, end=1, pan, accelerate, n, voice=0, decay=1 |
+	var dur = (end-begin)/speed;
+	var basefreq = n.midicps;
+	// lowest modes for clamped circular plate
+	var freqlist =[1.000,  2.081,  3.414,  3.893,  4.995,  5.954,  6.819,  8.280,  8.722,  8.882, 10.868, 11.180, 11.754,
+		13.710, 13.715, 15.057, 15.484, 16.469, 16.817, 18.628]**1.0;
+	var tscale = 100.0 / basefreq / (freqlist**(2-clip(decay,0,2)));
+	var ascale =freqlist**clip(voice,0,4);
+	var sound = Mix.arFill(15, {arg i; EnvGen.ar(Env.perc(0.01*tscale[i], 0.5*tscale[i], 0.2*ascale[i] ), timeScale:dur*5)
+		* SinOsc.ar(basefreq * freqlist[i] * Line.kr(1, 1+accelerate))});
```
telephon: maybe `Line.kr(1, 1+accelerate, sustain)` ? or `Line.kr(1, 1+accelerate, 1/speed)` ?

synths/default-synths.scd
```diff
@@ -75,3 +75,197 @@ SynthDef(\gabor, { |out, sustain = 0.03, note = 0, amp = 1, pan = 0, width = 0.3
...
+	sound = sound + (LPF.ar(WhiteNoise.ar(1), 1500*pitch1*Line.kr(1,1+accelerate,clickdur) ) * Line.ar(1, 0, clickdur));
+	OffsetOut.ar(out, DirtPan.ar(sound, ~dirt.numChannels, pan, env))
+}).add;
+
+// A vaguely 808-ish kick drum
+// "n" controls the chirp frequency, "speed" the filter sweep speed, and "voice" the sinewave feedback
+SynthDef(\super808, {|out, speed=1, begin=0, end=1, pan, voice=0, n |
+	var env, sound, dur, freq;
+	dur = 1.0 * (end - begin)/speed;
+	n = ((n>0)*n) + ((n<1)*3);
```
telephon, line 201: `n = ((n>0)*n) + ((n<1)*3);` (just curious why this is done!)

telephon:
I actually should fix the `gabor grain` example, using freq, but I'll wait until this one is merged.

synths/default-synths.scd
bgold-cosmos: OK, I didn't realize that.  I think that'll simplify the examples, so I'll change them.  I overload `speed` sometimes, but I hardly ever do anything with `begin` or `end`.

synths/default-synths.scd
bgold-cosmos, line 137: It's a decay time, but in the initial driving impulse rather than the sound itself.  And I just arrived at it rather arbitrarily after some experimentation.

synths/default-synths.scd
bgold-cosmos: I could try that - my guiding thoughts were to a) reuse existing Tidal parameters where it seemed to make sense and b) keep "useful" ranges in the [0,1] interval, since that's typical Tidal usage.  But frequencies are an odd case.

synths/default-synths.scd
bgold-cosmos: Good catch, I'll try that.

synths/default-synths.scd
bgold-cosmos, line 201: Here I'm using `n` more like a sample index.  Originally I just had `n.midicps`, but having all possible available pitches available for the drum seemed overkill, and what sounded best to me was around `30.midicps`.  So I put in an awful hack to have n=0 correspond to 30.midicps so the default `super808` would sound good, and other `n` values skip notes 10 at a time.

bgold-cosmos:
A general question that I had when writing these: Is there a way (using `SynthDescLib` maybe?) to ask SuperCollider either a) what SynthDefs have been defined or b) what arguments a given SynthDef can take?

telephon:
This shows how to do it:

```
x = SynthDescLib.global.synthDescs;
x.keysValuesDo { |key, val|
    key.postln;
    val.controlNames.postln;
    val.controls.collect { |x| x.defaultValue }.postln;
    "----------".postln;
};
```

How you would like to use it?

synths/default-synths.scd
telephon: Ah good, yes. All parameters can be used in any combination. It's good to boil the default-synths down to the simplest case so others can use them to build their own from them.

bgold-cosmos:
Exactly what I wanted!  At the very least it's nice for automated documentation, where I can make a list of all the synths I've added and what parameters they accept.  I imagine if it were wrapped in a nice function, end users could use it too (maybe just the part about querying controlNames).

I'm trying to keep to a fairly small set of "standard" parameters, but I find some do or don't make sense depending on the synth.  I don't imagine it's very obvious to end users that `superhat` makes use of `sustain, pan, accelerate, n` while `superclap` uses `speed, sustain, pan, n, delay, pitch1`, and I'm not sure how to be most helpful about it.

yaxu:
Yes there are many parameters defined in tidal already, should we be putting more effort into standardising them?
Or should tidal be able to query the available parameters and automatically create accessors for them somehow?

telephon:
> Or should tidal be able to query the available parameters and automatically create accessors for them somehow?

That would be very good. SuperDirt could send an OSC message to tidal with a list of parameters to add, maybe?

And yes, some standard would be useful.

yaxu:
Yes, I'm not sure how this would work in practice though, it's not too usual for a haskell program to make its own functions..

telephon:
What about allowing any keyword as parameter name and just letting dirt do the checking/ignoring?

bgold-cosmos:
Haskell doesn't really let you introspect on names of functions, so I think the closest you could get would be a syntax something like `d1 $ sound "superpiano*8" # paramF "muffle" "1 0.5 0.2 0"`.

But what I'm thinking about is more the issue of "how does the user know `muffle` even exists in the first place?".  I can certainly document it, but then there's a list of parameters for every single instrument...

telephon:
Couldn't then get even closer syntactically? E.g.

`let (##) op str = (#) paramF op str`

then you could write `d1 $ sound "superpiano*8" ## "muffle" "1 0.5 0.2 0"`

To know which work, you just have a look at the source code, or you query the list (like the example I gave above).

telephon:
@bgold-cosmos looks good now – can I merge?

bgold-cosmos:
I think OK to merge.  I've got some other synths I'm working on, but not near done and they can go in a separate PR.

I like the `##` idea, the only issue I can think of is if you later want to do `|+| muffle "1"` you don't have the `muffle` function available.  Might be something that doesn't really come up, and if it does `paramF` could still be there.

telephon:
ah yes, the combination of the two (`#` and `##`) doesn't work well. So in the end we'd need to have variants of `|+|` etc. as well. Or there could be a typeclass that combines `Pattern` and `(Pattern -> Pattern)` (I'm not sure if that is correct) that would allow a more general version of `|+|` and the like, so it could take either `"muffle"` or `muffle`?

PR #15, opened 2016-06-02
more musical synths

bgold-cosmos:
A total of six new synths. four analog-style (`supersquare`,
`supersaw`, `superpwm` and `supercomparator`) and two digital
(`superchip` and `supernoise`).

These make use of some of the VolcaKeys params like `voice`,
`semitone`, `lfo`, etc...  Documentation is in the comments.

### DIFF
--- a/synths/default-synths.scd
+++ b/synths/default-synths.scd
@@ -101,23 +101,26 @@ SynthDef(\tutorial2, {|out, sustain, pan |
 }).add;
 
 // it's also nice to control the pitch
-// using midicps means the "n" parameter will be used as MIDI note number
+// the "n" parameter will be used as MIDI note number and converted to "freq" by SuperDirt
 // also, "accelerate" will cause the pitch to drift
-// in Tidal we'll need to say something like `sound "tutorial3:69"` to hear a reasonably high pitch
-SynthDef(\tutorial3, {|out, sustain, pan, accelerate, n |
+// in Tidal we'll need to say something like `sound "tutorial3:69"`(440 Hz, 69 is a5 in Tidal notation)
+// to hear a reasonably high pitch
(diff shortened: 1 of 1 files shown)


synths/default-synths.scd
```diff
@@ -253,4 +253,130 @@ SynthDef(\supersiren, {|out, sustain, pan, n |
...
+// lfo(1) - how much the LFO affects the filter frequency
+// resonance(0.2) - filter resonance
+// voice(0.5) - depends on the individual synth
+
+// a moog-inspired square-wave synth; variable-width pulses with filter frequency modulated by an LFO
+// "voice" controls the pulse width (exactly zero or one will make no sound)
+SynthDef(\supersquare, {|out, speed=1, decay=0, sustain, pan, accelerate, n,
+	   voice=0.5, semitone=12, resonance=0.2, lfo=1, pitch1=1|
+	var env = EnvGen.ar(Env.pairs([[0,0],[0.05,1],[0.2,1-decay],[0.95,1-decay],[1,0]], -3), timeScale:sustain, doneAction:2);
+	var basefreq = n.midicps * Line.kr(1, 1+accelerate, sustain);
```
telephon: instead of using `n`, you could use the new `freq` parameter to calculate the basefreq:

`freq * Line.kr(1, 1+accelerate, sustain);`

Then if someone changes the pitch calculation in superdirt, it will be used by this synth.

telephon:
excellent, thank you!

synths/default-synths.scd
bgold-cosmos: OK, but it looks like `freq` defaults to `n+60`?  I use Tidal's midi notation parsing, so that results in awfully high pitches...

telephon: OK, I had in mind that `n = 0` would mean midinote 60. What is better to change?

PR #16, opened 2016-06-02
Topic performance improvements

telephon:
I'm not sure if we should merge these before making 0.8-dev public.


telephon:
(merged)

PR #21, opened 2016-06-04
conditional load of supplied sample path

sfradkin:
adding conditional around loading supplied path so that default samples are always loaded and it checks to see if a path was supplied before attempting to load from that path

### DIFF
--- a/classes/extSuperDirtStartup.sc
+++ b/classes/extSuperDirtStartup.sc
@@ -14,7 +14,10 @@
 		// boot the server and start SuperDirt
 		server.waitForBoot {
 			~dirt = SuperDirt(numChannels, server);
-			~dirt.loadSoundFiles(path);   // load samples (path can be passed in)
+			"loading default samples".postln;
+			~dirt.loadSoundFiles; // load default samples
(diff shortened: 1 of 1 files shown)


classes/extSuperDirtStartup.sc
```diff
@@ -14,7 +14,10 @@
 		// boot the server and start SuperDirt
 		server.waitForBoot {
 			~dirt = SuperDirt(numChannels, server);
-			~dirt.loadSoundFiles(path);   // load samples (path can be passed in)
```
telephon, line 17: thank you! I did this this morning, it should already be in master here: https://github.com/musikinformatik/SuperDirt/commit/f96f5fcfe69716c3595726e72d495f8c0fff2580

sfradkin, line 17: Yes. What I found was that when sending a path into SuperDirt.start, it would only load the samples on the path I supplied and did not load the default samples as well.  My change here ensures that the default samples are loaded whether or not a user supplies another path.

telephon:
(done, but not hardcoded).

PR #25, opened 2016-07-27
0.9 accelfix

bgold-cosmos:
Changes the `accelerate` parameter to produce a linear (in time) change
to speed.  This is more in line with the parameter's behavior in
classic Dirt.

### DIFF
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -84,8 +84,9 @@ DirtEvent {
 
 		if (~unit == \c) { speed = speed * ~unitDuration * ~cps };
 
-		endSpeed = speed * (1.0 + (accelerate.abs.linexp(0.01, 4, 0.001, 20, nil) * accelerate.sign));
-		if(endSpeed.sign != speed.sign) { endSpeed = 0.0 }; // never turn back
+		// endSpeed = speed * (1.0 + (accelerate.abs.linexp(0.01, 4, 0.001, 20, nil) * accelerate.sign));
+		// if(endSpeed.sign != speed.sign) { endSpeed = 0.0 }; // never turn back
+		endSpeed = speed * (1.0 + accelerate);
 		avgSpeed = speed.abs + endSpeed.abs * 0.5;
 
 		if(~unit == \rate) { ~unit = \r }; // API adaption to tidal output
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -84,8 +84,6 @@ DirtEvent {
 
 		if (~unit == \c) { speed = speed * ~unitDuration * ~cps };
 
-		// endSpeed = speed * (1.0 + (accelerate.abs.linexp(0.01, 4, 0.001, 20, nil) * accelerate.sign));
-		// if(endSpeed.sign != speed.sign) { endSpeed = 0.0 }; // never turn back
 		endSpeed = speed * (1.0 + accelerate);
 		avgSpeed = speed.abs + endSpeed.abs * 0.5;
 


classes/DirtEvent.sc
```diff
@@ -84,8 +84,9 @@ DirtEvent {
 
 		if (~unit == \c) { speed = speed * ~unitDuration * ~cps };
 
-		endSpeed = speed * (1.0 + (accelerate.abs.linexp(0.01, 4, 0.001, 20, nil) * accelerate.sign));
-		if(endSpeed.sign != speed.sign) { endSpeed = 0.0 }; // never turn back
```
telephon, line 88: Is there a reason to remove this line? Is it a feature to turn back?

classes/DirtEvent.sc
```diff
@@ -84,8 +84,9 @@ DirtEvent {
...
+		// endSpeed = speed * (1.0 + (accelerate.abs.linexp(0.01, 4, 0.001, 20, nil) * accelerate.sign));
+		// if(endSpeed.sign != speed.sign) { endSpeed = 0.0 }; // never turn back
+		endSpeed = speed * (1.0 + accelerate);
```
telephon, line 87: In a way, the accelerate parameter has an inconvenient asymmetry:
- when it is 1, it'll double the speed by the end
- when it is -1/2, it'll half the speed by the end
- when it is -1, it'll go to zero

Somehow it would be clearer if it were a ratio, defaulting to 1 for no acceleration, 2 for doubling and 1/2 for halving.

bgold-cosmos, line 87: I would find `accelerate "1"` == no acceleration to be very bizarre.

The "classic" definition is that for every sample, playback speed is increased by accelerate/samplerate.  So `accelerate "1"` means `speed` increases at one unit per second (for the default `unit "r"`).  Maybe it's best to recreate that?

telephon, line 87: yep, in terms of terminology, this is right. What we could do is to supply two parameters:
- `accelerate` (default 0): delta-speed (either relative to cps or not)
- `hurry` (default 1): endSpeed derived from speed

then we can define in SuperDirt:

```
endSpeed =  speed * hurry * (accelerate + 1)
```

It's a bit like `harmonic` and `detune` for frequencies.

One can define in the default event:

```
~hurry = { ~accelerate + 1 };
```

and then just do:

```
endSpeed =  speed * ~hurry.value

```

So hurry would override accelerate.

Well, just an idea, let me know what you think.

bgold-cosmos, line 87: Does it need to be done in SuperDirt?  With those definitions it looks like the relation between `hurry` and `accelerate` is simply (in Tidal) `let hurry pat = accelerate (fmap (subtract 1) pat)`

telephon, line 87: ah so you can add that to Param.hs and then write `d1 $ s "can" # hurry "2"`?

bgold-cosmos, line 87: Yeah, or add it anywhere, really.  `up` is sort of the same idea (an alternate way of expressing `speed`), and that's defined in Strategies.hs

telephon, line 87: good. What is a bit trickier is to make accelerate depend on dur. This is because the dur itself is defined in terms of speed and accelerate. It requires solving a quadratic equation, I think, so not too bad, but would need to be done with a bit of extra care to actually work for all the `unit`cases. I'm not sure it's worth the trouble ...

bgold-cosmos, line 87: Well, by "depending on dur" I really meant only because SuperDirt implements the effect by calculating a target `endspeed` (independent of dur), and essentially works out the dur and \Delta speed per time to get there.  Classic Dirt instead calculates the \Delta speed per time (independent of dur) straight from `accelerate`, since it's barely even aware of dur.

In the end, given that `unit` throws a wrench into the whole thing anyway, I'm not sure that it matters too much precisely what the implementation is, as long as it "sounds right".  And my opinion is that constant Delta speed per time is the most relevant thing to "sounds right". `accelerate`'s interactions with `speed`, `unit`, `sustain`, `end`, etc, are probably not something most users have intuition or expectations for.

tldr; it's a dev version - I support keeping the implementation simple, let the interactions fall where they do and fix later if something super weird crops up

telephon:
… so I suppose you agree that it's best to merge your PR now and if necessary care about the independence from duration later if necessary?

bgold-cosmos:
I think so.  This is a dev branch, it doesn't obviously break anything, and can be tweaked later if something seems off.

telephon:
thanks!

PR #30, opened 2016-11-25
Do not wrap orbits by default

telephon:
Do not wrap orbits by default.

It may be confusing that setting an orbit higher than the number of existing ones, it wraps and influences the default one (`0`).

You can switch on the wrapping if needed: `~dirt.wrapOrbits = true`.

### DIFF
--- a/classes/SuperDirt.sc
+++ b/classes/SuperDirt.sc
@@ -19,7 +19,7 @@ SuperDirt {
 
 	var <port, <senderAddr, <replyAddr, netResponders;
 	var <>fileExtensions = #["wav", "aif", "aiff", "aifc"];
-	var <>verbose = false, <>maxLatency = 42;
+	var <>verbose = false, <>wrapOrbits = false, <>maxLatency = 42;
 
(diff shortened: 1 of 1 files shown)


telephon:
@yaxu and all: any opinions on this PR? Is it better to wrap (which makes some patterns easier to write) or to warn (which avoids misunderstandings).

yaxu:
I'm not clear on the exact options here but I think it would be best to both wrap (take the modulo of number of orbits, because during a performance that's probably the best assumption) and warn (because I think that would rarely be intended behaviour). Relatedly, is there a way to stop floods of repeated warnings?

telephon:
Yes, this PR doesn't stop your performance, it would just warn if you run over.

>Relatedly, is there a way to stop floods of repeated warnings?

Well, good question. How is the bevaviour best described? You may want to be reminded that it repeats, no?

telephon:
Ah I see, it should actually wrap and warn.

telephon:
Committed a solution outside this conflicting PR here.

PR #40, opened 2016-12-27
dirt: new panning scheme

telephon:
This is a new panning scheme, where in multiple input channels are
distributed (“splayed”) across multiple output channels.

The conversion unipolar -> bipolar in the `pan` parameter is now done
in sclang, which is more efficient.

This is an API change. The defaultMixingFunction is replaced by a
defaultPanningFunction, so that the full panning behaviour can be
changed if needed.


telephon:
Don't merge yet, its still to be tested.

PR #42, opened 2017-01-09
Tremolo audio DSP effect creation.

pd3v:
Tremolo audio DSP effect declaration.

### DIFF
--- a/synths/core-synths.scd
+++ b/synths/core-synths.scd
@@ -284,6 +284,18 @@ live coding them requires that you have your SuperDirt instance in an environmen
 
 	}, [\ir, \ir]).add;
 
+	// Tremolo audio DSP effect definition. Params: tremolorate (Hz); tremolodepth (amplitude)
+	SynthDef("tremolo" ++ numChannels, { |out, tremolorate = 1.0, tremolodepth = 0.5 |
+		var signal, tremolowave;
+
+		tremolodepth = tremolodepth * 0.5;
+		tremolowave = (1 - tremolodepth) + SinOsc.ar(tremolorate, 0, tremolodepth);
+
+		signal = In.ar(out, numChannels);
+		signal = signal * tremolowave;
+
+		ReplaceOut.ar(out, signal);
+	}, [\ir, \ir, \ir]).add;
 
 	"---- core synth defs loaded ----".postln;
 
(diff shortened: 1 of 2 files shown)


synths/core-synths.scd
```diff
@@ -284,6 +284,18 @@ live coding them requires that you have your SuperDirt instance in an environmen
 
 	}, [\ir, \ir]).add;
 
+	// Tremolo audio DSP effect definition. Params: tremolorate (Hz); tremolodepth (amplitude)
+	SynthDef("tremolo" ++ numChannels, { |out, tremolorate = 1.0, tremolodepth = 0.5 |
+		var signal, tremolowave;
+
+		tremolodepth = tremolodepth * 0.5;
+		tremolowave = (1 - tremolodepth) + SinOsc.ar(tremolorate, 0, tremolodepth);
```
telephon: this is easier written as: `SinOsc.ar(tremolorate).range(0, tremolodepth)`. Another thing, when the rate is very low, this will mute the sound, because the sine wave starts at `0`. So you could write: `SinOsc.ar(tremolorate, 0.5pi).range(0, tremolodepth)` to make it start at full level.

telephon: … ups sorry, ignore the first comment.

you want the tremolowave to range from `1 - tremolodepth` to `1 + tremolodepth`. So whatever! Maybe easiest is: `.range(1 - tremolodepth, 1 + tremolodepth)`.

telephon: (edited)

When the rate is very low, this will mute the sound, because the sine wave starts at `0`. So you could write: `SinOsc.ar(tremolorate, 0.5pi)` to make it start at full level.

also,  you may want to limit it: `tremolodepth = tremolodepth.clip(0, 1)`

pd3v:
@telephon, I changed the code to clip both on **tremolorate** and **tremolodepth** params, as you suggested. More solid now.

telephon:
@pd3v thank you, this looks very good. If I were you, I would now do the following: in your branch, do a `git reset --soft master/0.9-dev`. Then if you do `git diff` you should just see your change. Then make a new branch with `git checkout -b topic-tremolo`, push that and make a new pull request.

It is fine as it is, but looks much nicer in the history ...

pd3v:
@telephon, `git reset --soft master/0.9-dev` gives back the following:

>  fatal: ambiguous argument 'master/0.9-dev': unknown revision or path not in the working tree.

`git status` gives back:

> 'origin/0.9-dev'

Should I use this instead?

telephon:
>Should I use this instead?

Yes, I think that should work, but I am not completely sure.

PR #43, opened 2017-01-11
Tremolo effect audio DSP.

pd3v:


pd3v:
Thank you @telephon .If I understand it right, when making a pull-request I should first create a branch (topic-_something_) and avoid work it directly from 0.9-dev? There's a conflict in **synths/core-synths.scd**?!

PR #47, opened 2017-01-24
Phaser audio DSP effect creation.

pd3v:


synths/core-synths.scd
```diff
@@ -314,6 +314,30 @@ live coding them requires that you have your SuperDirt instance in an environmen
...
 		ReplaceOut.ar(out, signal);
 	}, [\ir, \ir, \ir]).add;
+	
+	// Phaser audio DSP effect definition. Params: phaserrate (Hz); phaserdepth (amplitude)
+	SynthDef("dirt_phaser" ++ numChannels, { |out, phaserrate = 1.0, phaserdepth = 0.5 |
+		var signal, phaserwave;
+		var maxDelayTime = 4;
+		var phase =  LFPar.kr(phaserrate).range(0.0088, 0.01);
+
+		~phaserBus = Bus.audio(s, 2);
```
telephon: This will break if you have several instances of SuperDirt running.

synths/core-synths.scd
```diff
@@ -314,6 +314,30 @@ live coding them requires that you have your SuperDirt instance in an environmen
...
+		phaserdepth = phaserdepth.clip(0, 1);
+
+		signal = In.ar(out, numChannels);
+		Out.ar(~phaserBus, signal);
+
+		// phaser algorithm #1
+		phaserwave = AllpassL.ar(In.ar(~phaserBus), maxDelayTime, phase, 0, phaserdepth);
```
telephon: Why you need `~phaserBus`? It seems you are just feeding the signal straight back into the synth. Or do you need a global bus to mix all the signals of all running synths?

telephon (changes requested):

synths/core-synths.scd
pd3v: To have different phasers for different SuperDirt connections. 
For instance, d1 and d2 running two different phasers with different params values.

synths/core-synths.scd
pd3v: Hi @telephon! I don't know why. Let me check that. Thanks.

telephon: No need to test, I know it from reading the code. Unless you specifically prevent this, two instances of SuperDirt will run in the same environment. Using dynamic scope (`~something`) you will override them. Normally, unless in spontaneous hacks, you don't really use such constructions in a `SynthDef`.

synths/core-synths.scd
telephon: Doing it this way won't work out. For allocating resources-per-orbit, we still need a scheme. But the good news is: you don't need to allocate those busses at all, because each orbit has its own synth and an `AllpassL` has its own memory.

telephon:
thanks! Could you be so kind and do another PR based on the latest 0.9-dev? This one is a bit complicated to read … (sorry, these git things can sometimes be a little uncomfortable)

PR #57, opened 2017-03-07
Phaser audio dsp effect creation

pd3v:
@telephon, new PR to replace previous one. Better now.

### DIFF
--- a/synths/core-synths.scd
+++ b/synths/core-synths.scd
@@ -330,16 +330,15 @@ live coding them requires that you have your SuperDirt instance in an environmen
 	}, [\ir, \ir, \ir]).add;
 
 	// Phaser audio DSP effect definition. Params: phaserrate (Hz); phaserdepth (amplitude)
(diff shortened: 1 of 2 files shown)


synths/core-synths.scd
```diff
@@ -329,6 +329,24 @@ live coding them requires that you have your SuperDirt instance in an environmen
 		ReplaceOut.ar(out, signal);
 	}, [\ir, \ir, \ir]).add;
 
+	// Phaser audio DSP effect definition. Params: phaserrate (Hz); phaserdepth (amplitude)
+	SynthDef("dirt_phaser" ++ numChannels, { |out, phaserrate = 1.0, phaserdepth = 0.5 |
+		var signal, phaserwave;
+		var maxDelayTime = 4;
```
telephon: you can do one thing: add `sustain` as an argument to the synth, and use that value as a `maxDelayTime`. Saves *a lot* of memory!

telephon (changes requested):
looks good! just one little thing

telephon (changes requested):
that was a misunderstanding: the Parameter name must be 'sustain' to work automatically. thanks !!

PR #58, opened 2017-03-31
hash calculation adjustment

bgold-cosmos:
Hash on the buffer, rather than `sound`.  This affects the behavior of
`cut -1`, making it so only identical samples cut each other (where
"identical" means loaded from the same file).  This follows the
documented behavior and how Classic Dirt treated the parameter.

### DIFF
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -36,7 +36,6 @@ DirtEvent {
 		var buffer, sound, synthDesc, sustainControl;
 		sound = ~s;
 		buffer = orbit.dirt.getBuffer(sound, ~n);
-		~hash = ~hash ?? { buffer.identityHash };
 
 		if(buffer.notNil) {
 			if(buffer.sampleRate.isNil) {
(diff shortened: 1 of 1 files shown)


telephon:
Thank you! But the way you do it here will break cut for synths.

Probably a better way is to make the hash on `~s` and `~n` together:

`~s.identityHash bitXor: ~n.identityHash`

telephon:
… or wait, for synths, `n` shouldn't play a role. So better make two ways to produce the hash, after the branch further below.

bgold-cosmos:
I wasn't sure how `cut "-1"` really should work for synths.  Hashing on the buffer seemed to work for synths, but maybe it's buggy.

Actually, I'm a bit fuzzy on `identityHash`, I don't quite get how it works for `cut`.  Why doesn't `~s` wind up with a new hash every time it gets changed?  If I take two variables `a = "foo"` and `b = "foo"` they have different hashes...

telephon:
`~s` is a symbol - that has identity. (`\foo === \foo`).

bgold-cosmos:
Ah, I had been thinking of `\foo` as just convenience for `"foo"`, but they're not the same.

I'll redo the hash calculation in the branches...

telephon:
cool, I'll do it for 1.0-dev

telephon:
Ah @bgold-cosmos I just saw your commit was for 1.0-dev - I have a different solution here. But I'll merge this into master, where it should go now.

PR #60, opened 2017-05-10
Update missing sample

msp:
Updated whilst testing it out.


telephon:
Thanks, I had overlooked your PR.

PR #72, opened 2018-02-03
use speed for note and freq

telephon:
You might want to use `speed` separately from `note` in softsynths, but since there are a lot of functions in tidal that will use `speed` to control pitch (in samples) it might become annoying that this isn't the same for synths.

Feel free to disagree.

### DIFF
--- a/classes/SuperDirt.sc
+++ b/classes/SuperDirt.sc
@@ -494,7 +494,7 @@ DirtOrbit {
 			~n = 0; // sample number or note
 			~octave = 5;
 			~midinote = #{ ~n + (~octave * 12) };
-			~freq = #{ ~midinote.midicps };
+			~freq = #{ ~midinote.midicps * ~speed };
 			~delta = 1.0;
 
 			~latency = 0.0;


yaxu:
This is a good idea I think, but what about negative speeds -- should that invert the envelope over time?

bgold-cosmos:
My first reaction was rather ambivalent - I've gotten used to the current behavior.  But I suspect in the long run it's the right thing to do (principle of least surprise).

Some of the synths in default-synths-extra.scd overload speed to do stuff with filter envelopes or whatever, but those could be rewritten if necessary.  Someone (probably me) should take a second look at those anyway - quite a bit of the SuperDirt "API" has changed since they were written.

telephon:
yes, I also hesitate, because you may want to use `speed` in a different way, and also `accelerate` isn't so simple to implement outside of a synth (not impossible though).

The basic problem is that we use `n` for two different things in tidal. We could do the following: map `n` to whatever it is now. But map `note` to `speed` for samples, so that `up` etc. can map to `note`.

yaxu:
Would `note` and `n` be the same for synths?
From another perspective maybe `n` means exactly the same for synths and samples (play a particular sound from a family `s` of related sounds), it's just that some `s` (both samplebanks and synths) order the sounds in a particular way by convention. The fundamental problem could then be that this convention is not well defined and conflicts with other parameters. i.e. the relationship between pitch and timbre is unresolved.

yaxu:
I think we talked before about renaming `speed` to`rate`. Would that help at all?

telephon:
>From another perspective maybe n means exactly the same for synths and samples (play a particular sound from a family s of related sounds),

Yes, here is what currently happens: when a synth has a `freq` parameter, it draws from the following functions (they are customisable, but this is the default):

```supercollider
~midinote = #{ ~n + (~octave * 12) };
~freq = #{ ~midinote.midicps };
```

The parameter `n` is used _both_ to select a variant (sample or synth if it exists), and, for synths, to determine the frequency. The reason is that you want to use scales etc. in tidal and they determine `n`. If these functions would determine `note`, `detune`,   or `harmonic`, they could be used for both synths and samples (for which we could assume some base frequency, or specify it).

The question is how best to control the frequency of both synths and samples in the same way, so that all frequency/speed related functions in tidal map to both.

Issue #74, opened 2018-02-10
Editor integration


yaxu:
Two ideas really, but thought I'd put them together as maybe general discussion about editor integration would be nice
First is, shall we create ten or eleven orbits by default, and have each connection (d1..) go to a different orbit (again by default)?
Second, would it be possible for an editor to subscribe to an rms analysis of each orbit? For display in the editor next to active patterns

telephon:
yes, why not. 

for the first: we just change the startup file, and add default orbit args in tidal? One thing to worry about (or rather: to test) is if this might lead to a very large cpu load when you use many at the same time, all with effects. also, it is an aesthetic choice. it sounds very differently.

rms: yes. can we open a port in tidal that dirt can send to?

yaxu:
Yes default orbit args would be added to the editor startup config. I'll experiment with this and see what the cpu load is like.

(Currently I'd say global effects don't work well when multiple patterns send to the same orbit. I haven't investigated in detail but think when a pattern doesn't set a parameter, it turns off the effect.)

rms - I think it wouldn't be tidal itself that would open the port, but the editor. I'm working on a new text editor in ncurses, so the vu meter would be rendered with extended ascii / unicode blocks, so wouldn't need to be high resolution in this case! FFT might also be nice to experiment with. I'd want to mirror the functionality in classic dirt.

fdragovic:
either they get turned off, although it sounds like it's just the default
values that are accessed because delays and reverbs jump back and forth
between values... would be really cool if there was a way around that!

also, ncurses editor sounds teh sick :D

telephon:
yes in the end it just depends on what kind of orbit-effects you use.

telephon:
if we would be picky, we would say that a delay is local, but a room is global. we would need nested orbits...

fdragovic:
sounds like that would be another cpu issue though seeing as those global
effects are always on standby? maybe there is a way to ignore the defaults
if there is something already sending values on that orbit?

echophon:
Would it make sense to have an orbit or two dedicated for global fx? With
individual dirt channels using a send 0-1.0 value to feed the signal into
the effect.

telephon:
@fdragovic if you don't use them, they don't eat. But if you do, each parallel dirt pattern will use one each instead of one total.

bgold-cosmos:
[...]

PR #76, opened 2018-02-14
rms analysis support

telephon:
This branch adds rms analysis, it sends data to all registered clients of the server. The format is:

`["/rms", synthID, orbitIndex, peak1, power1, peak2, power2,  …]` (continuing for all audio channels).

The sending can be activated by sending the parameters:

`# rmsPeakLag "3" # rmsReplyRate "8"`  from tidal (you need to add those to tidal).


telephon:
see: #74

yaxu:
Thanks @telephon,
Pardon my ignorance, but how would the editor become a registered client of the server? Would doing this mean getting other OSC messages intended for sclang?

Presumably there's no issue with 'hand crafting' an OSC message to set the `rmsPeakLag` and `rmsReplyRate`, rather than going via Tidal?

fdragovic:
thanks! trying to learn how the effects chains are handled and this and the
global effects pr are really helping :D

On Feb 14, 2018 3:56 PM, "Julian Rohrhuber" <notifications@github.com>
wrote:

see: #74 <https://github.com/musikinformatik/SuperDirt/issues/74>

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/musikinformatik/SuperDirt/pull/76#issuecomment-365632196>,
or mute the thread

fdragovic:
@yaxu it should already be sending osc just to 57110. if you filter out the
messages on that port instead of 6010 you can get the ones it sends to
sclang directly i think (:

yaxu:
Sorry @fdragovic not sure exactly what you mean.. Ideally I just want the rms messages, going to port 6010.

telephon:
you just send an OSC message `"/notify"` to scsynth, that is, normally to port `57110`.

telephon:
sorry, I didn't mention this. I was a bit lazy and just followed the standard scheme for supercollider here.

If it helps, we can also route the messages through sclang and send exactly where you want to have it.

But maybe it is also easy like this. The editor must know the server port of course, which might be a disadvantage.

This is a general way to send information from the server to tidal or any other application.

yaxu:
How do you tell scsynth which port to notify? Maybe routing via sclang is indeed best..

telephon:
scsynth will know automatically from the message you send (there is a reply address in each)

yaxu:
How? As I understand it, the outbound port won't be the same port as the one that is being listened to.

yaxu:
Ah, I found this issue. https://github.com/supercollider/supercollider/issues/801
I've actually never seen a UDP client listening on the same port that they send from before. I'll look into doing this with Haskell.

telephon:
Ah yes, of course. I can add a rerouting via sclang if Haskell doesn't work.

telephon:
(doesn't send yet)

yaxu:
One more feature request for this - it would be nice to have the option to have left/right channels analysed separately, for stereo VU metering.

telephon:
This already happens (you get each of the channels separately, independent of the number of channels).

telephon:
ok, sends now. just call:

`~dirt.startSendRMS`

telephon:
again, the format is:

`["/rms", synthID, orbitIndex, peak1, power1, peak2, power2, …]` and so on for as many channels as you have specified.

telephon:
maybe the OSC command should be `/tidal_rms`.

yaxu:
I'm working on a minimal example of subscribing to scserver. Here's what I have so far, basically sending `/notify 1` - I see the `1` is needed from this: http://doc.sccode.org/Reference/Server-Command-Reference.html#/notify
```
import Sound.OSC.FD
import qualified Network.Socket as N
main = do udp <- udpServer "127.0.0.1" 0
          remote_addr <- N.inet_addr "127.0.0.1"
          let remote_sockaddr = N.SockAddrInet 57110 remote_addr
          sendTo udp (Message "/notify" [int32 1]) remote_sockaddr
          loop udp
  where loop udp = do m <- recvMessage udp
                      putStrLn $ show m
                      loop udp
```
But I get back `/fail "too many users"`

Quite a general issue but any clues appreciated!

telephon:
Can you try and start the server with more logins?
```
s.options.maxLogins = 8;
s.reboot;
```

yaxu:
@telephon Thanks, increasing the number of logins has indeed allowed more logins! :)

yaxu:
Ok I can see numbers now!
I have six orbits going to a single stereo output, i.e. `[0, 0, 0, 0, 0, 0]`

I put `~dirt.startSendRMS` at the end of my startup, and get this error repeated 11 times:
`ERROR: Meta_Bus:audio: failed to get an audio bus allocated. numChannels: 2 server: localhost`

However it seems to work anyway. I get one osc message per orbit, each with four float values, as you describe, and they look good.

To be honest I'm not clear on what the difference/relationship between `rms` and `peak` are, but will play around with them.. Classic dirt is only calculating rms so far.

Thanks a lot for your help with this.

telephon:
>I put `~dirt.startSendRMS` at the end of my startup, and get this error repeated 11 times:
`ERROR: Meta_Bus:audio: failed to get an audio bus allocated. numChannels: 2 server: localhost`

you may have to increase the 
`s.options.numAudioBusChannels = 1024 * 8`

But it shouldn't really happen (we don't use many busses). Maybe we you have an old version of one of the components (scynth /sclang).

>To be honest I'm not clear on what the difference/relationship between rms and peak are, but will play around with them.. Classic dirt is only calculating rms so far.

In the usual level indicators, you get a peak level of the signal as well, so that you can see if it has clipped. There is an example in `tests/testing-rms.scd`.

yaxu:
Increasing audiobuschannels seems to have fixed it, thanks. I'm on SC v3.8.1

This is working fine now, perhaps ready to be merged? The only thing is I found the default 8Hz too slow and upped to 15Hz. I suggest a default of at least 10.

Here's a quick demo:
  https://www.youtube.com/watch?v=l1MNP2MWGTI

(that's captured at 20Hz with rms at 15Hz, so Nyquist is among us..)

telephon:
> I'm on SC v3.8.1

That is too old now! https://github.com/supercollider/supercollider/releases

>The only thing is I found the default 8Hz too slow and upped to 15Hz. I suggest a default of at least 10.

The  defaults for the UGen are `replyRate = 20.0, peakLag = 3`. Should we use those?

PR #84, opened 2018-04-19
Monitor pause

bgold-cosmos:
The DirtPause UGen in dirt_monitor was causing SuperDirt to "fall asleep" if no audio was playing, and would take a few seconds to wake up.  Since it wasn't present before 1.0-dev, I've removed it.

### DIFF
--- a/synths/core-synths-global.scd
+++ b/synths/core-synths-global.scd
@@ -39,8 +39,6 @@ CORE SYNTHDEFS FOR DIRT
 		// or:
 		//signal = distort(signal);
 
-		// DirtPause.ar(signal, graceTime:4);
-
 		signal = signal * EnvGen.kr(Env.asr, gate, doneAction:2);
 		Out.ar(outBus, signal)
 	}, [\ir, \ir, \kr, \kr]).add;
--- a/synths/core-synths-global.scd
+++ b/synths/core-synths-global.scd
@@ -39,7 +39,7 @@ CORE SYNTHDEFS FOR DIRT
 		// or:
 		//signal = distort(signal);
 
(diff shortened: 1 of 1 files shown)


yaxu:
Thanks @bgold-cosmos, that does fix it for me. I've taken the liberty of doing the merge, hope that's OK

telephon:
Could you wait just a little more before merging? 1.0-dev is a very different beast, and you generally can't assume that things are just there for no reason. In this case, the pause in the monitor is there to allow you to have a large number of orbits without cost.

If the monitor takes time to "wake up", then there is some other problem that needs to be fixed, otherwise it'll bite us elsewhere.

telephon:
I've looked at the code (in `DirtOrbit`) and the synth should be started as soon as any OSC message comes in.

The relevant lines are:

https://github.com/musikinformatik/SuperDirt/blob/1.0-dev/classes/DirtOrbit.sc#L66

https://github.com/musikinformatik/SuperDirt/blob/1.0-dev/classes/GlobalDirtEffect.sc#L43

https://github.com/musikinformatik/SuperDirt/blob/1.0-dev/classes/GlobalDirtEffect.sc#L52

If you add some info post into the L52 above then you should see if it is called or not, e.g.

`if(someArgsNotNil) { "resuming global effect '%'".format(name).postln; synth.run };`

telephon:
I could reproduce it, trying to fix it now.

telephon:
I found the problem in `DirtPause`. Will need a little thought. Let's revert this merge, ok?

yaxu:
Yes sorry @telephon !

PR #101, opened 2018-12-08
Add TidalFX SynthDefs

calumgunn:
Adds effect SynthDefs from [TidalFX](https://github.com/calumgunn/TidalFX):

- [Waveloss](http://doc.sccode.org/Classes/WaveLoss.html)
- [Squiz](http://doc.sccode.org/Classes/Squiz.html)

Tested locally and all working fine - but the Tidal part isn't in place yet, so evaluate this in Tidal:
```
(squiz, _) = pF "squiz" (Just 1.0)
(waveloss, _) = pF "waveloss" (Just 0.0)
```

And use like so:
```
d1 $ sound "bd cp" # waveloss 40 # squiz 4
```

### DIFF
--- a/library/default-effects-extra.scd
+++ b/library/default-effects-extra.scd
@@ -1,0 +1,42 @@
+// Waveloss
+// Divides an audio stream into tiny segments, using the signal's
+// zero-crossings as segment boundaries, and discards a fraction of them.
+
+(
+SynthDef("waveloss" ++ ~dirt.numChannels, { |out, drop = 1|
+  var sig;
+
+  sig = In.ar(out, ~dirt.numChannels);
+  sig = WaveLoss.ar(sig, drop, outof: 100, mode: 2);
+  ReplaceOut.ar(out, sig)
+}).add;
+
+~dirt.addModule('waveloss', { |dirtEvent|
+  dirtEvent.sendSynth('waveloss' ++ ~dirt.numChannels,
+    [
+      drop: ~waveloss,
+      out: ~out
+    ]
+  )
+}, { true });
+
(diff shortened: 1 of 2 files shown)


library/default-effects-extra.scd
```diff
@@ -0,0 +1,42 @@
...
+}).add;
+
+~dirt.addModule('waveloss', { |dirtEvent|
+  dirtEvent.sendSynth('waveloss' ++ ~dirt.numChannels,
+    [
+      drop: ~waveloss,
+      out: ~out
+    ]
+  )
+}, { true });
```
telephon: should be:
`} , { ~waveloss.notNil }`

Otherwise, the module will be run on every synth.

library/default-effects-extra.scd
```diff
@@ -0,0 +1,42 @@
...
+}).add;
+
+~dirt.addModule('squiz', { |dirtEvent|
+  dirtEvent.sendSynth('squiz' ++ ~dirt.numChannels,
+    [
+      pitchratio: ~squiz,
+      out: ~out
+    ]
+  )
+}, { true });
```
telephon: same here with `~squiz`.

telephon (changes requested):
see comments …

calumgunn:
Ahhh, _that's_ what the boolean does! Of course. Updated.

telephon (approved):
I haven't tested it, but it reads good now!

yaxu:
I added the params to tidal (also untested)

calumgunn:
:) Thanks!

PR #113, opened 2019-02-21
Add SpectralTricks fx + 2 new spectral fx

madskjeldgaard:
This PR contains all of the fx synths from [SpectralTrics](https://github.com/madskjeldgaard/SpectralTricks) and adds two new spectral effects: conform and enhance. 

They all make use of custom parameters that I will add as a PR on the Tidal repo

### DIFF
--- a/library/default-effects-extra.scd
+++ b/library/default-effects-extra.scd
@@ -1,341 +1,383 @@
-// Waveloss
-// Divides an audio stream into tiny segments, using the signal's
-// zero-crossings as segment boundaries, and discards a fraction of them.
+/*
 
+DEFAULT EFFECTS EXTRA
+
+*/
 (
-SynthDef("waveloss" ++ ~dirt.numChannels, { |out, drop = 1|
-  var sig;
+	// Waveloss
+	// Divides an audio stream into tiny segments, using the signal's
+	// zero-crossings as segment boundaries, and discards a fraction of them.
 
-  sig = In.ar(out, ~dirt.numChannels);
-  sig = WaveLoss.ar(sig, drop, outof: 100, mode: 2);
-  ReplaceOut.ar(out, sig)
-}).add;
+	~dirt.addModule('waveloss', { |dirtEvent|
+		dirtEvent.sendSynth('waveloss' ++ ~dirt.numChannels,
+			[
+				drop: ~waveloss,
+				out: ~out
+			]
+		)
+	}, { ~waveloss.notNil });
 
-~dirt.addModule('waveloss', { |dirtEvent|
-  dirtEvent.sendSynth('waveloss' ++ ~dirt.numChannels,
-    [
-      drop: ~waveloss,
-      out: ~out
-    ]
-  )
-}, { ~waveloss.notNil });
+	SynthDef("waveloss" ++ ~dirt.numChannels, { |out, drop = 1|
+		var sig = In.ar(out, ~dirt.numChannels);
+		sig = WaveLoss.ar(sig, drop, outof: 100, mode: 2);
+		ReplaceOut.ar(out, sig)
+	},[\ir, \ir]).add;
 
-// Squiz
-// "reminiscent of some weird mixture of filter, ring-modulator
-// and pitch-shifter"
+	// Squiz
+	// "reminiscent of some weird mixture of filter, ring-modulator
+	// and pitch-shifter"
+	~dirt.addModule('squiz', { |dirtEvent|
+		dirtEvent.sendSynth('squiz' ++ ~dirt.numChannels,
+			[
+				pitchratio: ~squiz,
+				out: ~out
+			]
+		)
+	}, { ~squiz.notNil });
 
-SynthDef("squiz" ++ ~dirt.numChannels, { |out, pitchratio = 1|
-  var sig;
-  sig = In.ar(out, ~dirt.numChannels);
-  sig = Squiz.ar(sig, pitchratio);
-  ReplaceOut.ar(out, sig)
-}).add;
+	SynthDef("squiz" ++ ~dirt.numChannels, { |out, pitchratio = 1|
+		var sig = In.ar(out, ~dirt.numChannels);
+		sig = Squiz.ar(sig, pitchratio);
+		ReplaceOut.ar(out, sig)
+	}, [\ir, \ir]).add;
 
-~dirt.addModule('squiz', { |dirtEvent|
-  dirtEvent.sendSynth('squiz' ++ ~dirt.numChannels,
-    [
-      pitchratio: ~squiz,
-      out: ~out
-    ]
-  )
-}, { ~squiz.notNil });
+	// Frequency shifter
+	// Total shift is sum of `fshift` (in Hz) and `fshiftnote` times the current note frequency.
+	// `fshiftphase` allows control over the phase
+	~dirt.addModule('fshift', { |dirtEvent|
+		dirtEvent.sendSynth("dirt_fshift" ++ ~dirt.numChannels,
+			[
+				fshift: ~fshift,
+				fshiftphase: ~fshiftphase,
+				fshiftnote: ~fshiftnote,
+				freq: ~freq,
+				out: ~out
+			]
+		)
+	}, { ~fshift.notNil });
 
-// Frequency shifter
-// Total shift is sum of `fshift` (in Hz) and `fshiftnote` times the current note frequency.
-// `fshiftphase` allows control over the phase
-~dirt.addModule('fshift', {|dirtEvent| dirtEvent.sendSynth("dirt_fshift" ++ ~dirt.numChannels,
-	[fshift: ~fshift, fshiftphase: ~fshiftphase, fshiftnote: ~fshiftnote, freq: ~freq, out: ~out])}, {~fshift.notNil});
-SynthDef("dirt_fshift"++~dirt.numChannels, {|out, fshift, fshiftphase, fshiftnote, freq|
-	var sig = In.ar(out, ~dirt.numChannels);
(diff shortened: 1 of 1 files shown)


library/default-effects-extra.scd
```diff
@@ -120,4 +120,322 @@ SynthDef("dirt_distort" ++ ~dirt.numChannels, {|out, distort=0|
...
+
+*************************************/
+
+/****************************
+        SPECTRAL DELAY
+*****************************/
+
+~dirt.addModule('spectral-delay', { |dirtEvent|
+	dirtEvent.sendSynth('spectral-delay' ++ ~dirt.numChannels,
+		// OPTIONAL
```
telephon: It is also more efficient to explicitly specify the arguments. (maybe add this to your explanation, so we spread the news?)

library/default-effects-extra.scd
```diff
@@ -120,4 +120,322 @@ SynthDef("dirt_distort" ++ ~dirt.numChannels, {|out, distort=0|
...
+
+~dirt.addModule('spectral-freeze', { |dirtEvent|
+	dirtEvent.sendSynth('spectral-freeze' ++ ~dirt.numChannels,
+		[
+			freeze: ~freeze,
+			sustain: ~sustain,
+			out: ~out
+		]
+	)
+}, { ~freeze.notNil } ); // play synth only if at least one of the two was given
```
telephon: comment is now incorrect

library/default-effects-extra.scd
```diff
@@ -120,4 +120,322 @@ SynthDef("dirt_distort" ++ ~dirt.numChannels, {|out, distort=0|
...
+
+SynthDef("spectral-freeze" ++ ~dirt.numChannels, { |out, freeze , sustain|
+
+	var signal, chain, in;
+
+	signal = In.ar(out, ~dirt.numChannels);
+
```
telephon: could you reformat this SynthDef to match the standards? (ctrl-tab will do the most, and you can remove most empty lines)

library/default-effects-extra.scd
```diff
@@ -120,4 +120,322 @@ SynthDef("dirt_distort" ++ ~dirt.numChannels, {|out, distort=0|
...
+~dirt.addModule('spectral-comb', { |dirtEvent|
+	dirtEvent.sendSynth('spectral-comb' ++ ~dirt.numChannels,
+		[
+			comb: ~comb,
+			sustain: ~sustain,
+			out: ~out
+		]
+	)
+}, { ~comb.notNil}); // play synth only if at least one of the two was given
```
telephon: comment is incorrect now (also check all the ones below)

library/default-effects-extra.scd
```diff
@@ -120,4 +120,322 @@ SynthDef("dirt_distort" ++ ~dirt.numChannels, {|out, distort=0|
...
+
+SynthDef("spectral-comb" ++ ~dirt.numChannels, { |out, comb, sustain|
+
+	var signal, chain, in, clean, teeth=256;
+
+	signal = In.ar(out, ~dirt.numChannels);
+
```
telephon: same as above, formatting

library/default-effects-extra.scd
telephon: and, to make it perfect, add a space before the `}`, in all the ones below, too. Sorry that I am picky, but people will copy it and continue from this, so better do it now than later :)

library/default-effects-extra.scd
```diff
@@ -120,4 +120,322 @@ SynthDef("dirt_distort" ++ ~dirt.numChannels, {|out, distort=0|
...
+SynthDef("spectral-smear" ++ ~dirt.numChannels, { |out, smear, sustain|
+
+	var signal, chain, in;
+
+	signal = In.ar(out, ~dirt.numChannels);
+
+    chain = Array.fill(signal.size, {|i| FFT(LocalBuf(2048), signal[i])});
+
+    signal = IFFT(PV_MagSmear(chain, bins: smear*100));
```
telephon: is `100` a good scaling? Wouldn't you want something that is related to `2048` ?

library/default-effects-extra.scd
```diff
@@ -120,4 +120,322 @@ SynthDef("dirt_distort" ++ ~dirt.numChannels, {|out, distort=0|
...
+
+	var signal, chain, in, clean, teeth=256;
+
+	signal = In.ar(out, ~dirt.numChannels);
+
+    clean = signal;
+
+    chain = Array.fill(signal.size, {|i| FFT(LocalBuf(2048), signal[i])});
+
+    signal = IFFT(PV_BrickWall(chain, wipe:lbrick.range(0,-1)));
```
telephon: `range` only applies to UGens where we know the range. Because `lbrick` is a control, it could be anything. Do you want to clip the range? Then you can use `clip(0, 1)`. You could also wrap the range `wrap(0, 1)`, then any value will be significant in some way.

library/default-effects-extra.scd
```diff
@@ -120,4 +120,322 @@ SynthDef("dirt_distort" ++ ~dirt.numChannels, {|out, distort=0|
...
+
+	signal = In.ar(out, ~dirt.numChannels);
+
+    clean = signal;
+
+    chain = Array.fill(signal.size, {|i| FFT(LocalBuf(2048), signal[i])});
+
+	signal = IFFT(
+		PV_SpectralEnhance(chain,
+			partials.linlin(0.0,1.0, 1, 16),
```
telephon: formatting …

library/default-effects-extra.scd
```diff
@@ -120,4 +120,322 @@ SynthDef("dirt_distort" ++ ~dirt.numChannels, {|out, distort=0|
...
+	var signal, chain, in, clean, teeth=256;
+
+	signal = In.ar(out, ~dirt.numChannels);
+
+    clean = signal;
+
+    chain = Array.fill(signal.size, {|i| FFT(LocalBuf(2048), signal[i])});
+
+    signal = IFFT(
+		PV_ConformalMap(chain, real.linlin(0.0,1.0,0.01,2.0), imag.linlin(0.00,1.0,0.01,10.0))
```
telephon: formatting …

telephon (changes requested):
just minor things. Thanks a lot!

library/default-effects-extra.scd
madskjeldgaard: This is an old error of mine it seems. But the parameter needs to be scaled to the negative because that's when PV_BrickWall works as a low pass brick filter (positive values make it a high pass brick filter). But using range for this is wrong anyway so I'll correct this to linlin

library/default-effects-extra.scd
madskjeldgaard: 100 is the amount used in the help file examples. 2048 sounds really good but it uses waaaaay more cpu. In my testing, playing an arpeggio in Tidal with smear going from 0.0 to 1.0 uses around 1.8-3% cpu with 100 as scaling and 1.8-18% at 2048. That said I'll change this to an exponential scaling as it sounds better

madskjeldgaard:
Okay I changed my PR quite a bit now after looking at your thorough comments. I hope it conforms a bit better to the SuperDirt style now. Also I did a bit of rework and deeper testing of some of the effects to make them a bit more precise

madskjeldgaard:
Let me know how it looks / sounds

library/default-effects-extra.scd
telephon: you could also do `.clip2.neg`, but `linlin` might be clearer.

library/default-effects-extra.scd
telephon: ok, and maybe use 128 or 64 :) Or 108 :)

telephon:
Yes! very good.

Two more things to note:
- you are specifying `sustain` as a parameter everywhere, but you don't use it. You can remove it.
- adding the argument rates as a third argument of a SynthDef, typically `[\ir]`, saves a little bit of electricity.

A good file to look at for comparison is core-synths.scd (https://github.com/musikinformatik/SuperDirt/blob/develop/synths/core-synths.scd).

telephon:
I just saw that in default-effects-extra.scd there is a bit of formatting work to do. I can do it, but if you feel like it you could fix that as well ...

library/default-effects-extra.scd
madskjeldgaard: I honestly can't hear the difference between 100 and 128. 64 makes the effect a bit more subtle than 100/128 but also more effective. Let's make this 64 to be safe

madskjeldgaard:
I'll change those two things and I might as well look at formatting while I'm at it.

madskjeldgaard:
How do you want the comments ? The file you refer to use 
	```

	/*
	Effect SynthDefs
	These per-sample-effects are freed after Monitor envelope has ended
	*/
``` 

and the extra effects file uses
```
// Effect name
// explanation ...
```

madskjeldgaard:
Okay see latest commit where I revised the formatting throughout. Not 100% sure what to do with comments but at least it's fairly consistent within this file now. Also I think it's nice to have to add the module and then the synth def and keep that order throughout

madskjeldgaard:
Also, some of the synthdefs don't have default values in their arguments. Should these be set to something?

telephon:
Thank you, this is much better!!

Default arguments might be neat. If they are zero they can be left out.

You can leave the rates, I have an idea of how to simplify this.

madskjeldgaard:
Github says "Changes requested" but I'm too terrible at using Github to see anymore changes needed but let me know. Otherwise I'll consider this done

telephon:
Thanks!

madskjeldgaard:
No problem and thanks for all the comments!

PR #119, opened 2019-03-12
add a DJ filter

yaxu:


yaxu:
I added the `djf` control to tidal now

telephon (changes requested):
could you be so kind and check the code formatting? There seem to be some wrong indentations (use tabs). Also you can leave out the comment here.

library/default-effects-extra.scd
```diff
@@ -380,4 +380,37 @@ DEFAULT EFFECTS EXTRA
...
+	    	[
+	    		dj: ~dj,
+	    		out: ~out
+	    	]
+    	)
+    }, { ~dj.notNil});
+
+    SynthDef("dj-filter" ++ ~dirt.numChannels, { |out, dj|
+    	var signal;
+        var lpfCutoff = (dj*2) min: 1;
```
telephon: maybe these two would be easier to read if you write:

```
var lpfCutoff = dj.linlin(0, 0.5, 0, 1);
var hpfCutoff = dj.linlin(0.5, 1, 0, 1);
```

or you could do the mapping instantly:

```
var lpfCutoffFreq = dj.linexp(0, 0.5, 20, 10000);
var hpfCutoffFreq = dj.linexp(0.5, 1, 20, 10000);
```

then you can omit the `ControlSpec`.

yaxu:
I should say that this code came from @antonhornquist in the supercollider slack

PR #124, opened 2019-04-13
Add pan parameter to reverb effect

f-f:
Right now the reverb is always perfectly stereo even if the source is panned, so this PR adds the possibility of panning the reverb too

EDIT: I'm not sure if I opened the PR against the correct branch?

### DIFF
--- a/classes/DirtOrbit.sc
+++ b/classes/DirtOrbit.sc
@@ -60,7 +60,7 @@ DirtOrbit {
 		this.globalEffects = [
 			// all global effects sleep when the input is quiet for long enough and no parameters are set.
 			GlobalDirtEffect(\dirt_delay, [\delaytime, \delayfeedback, \delaySend, \delayAmp, \lock, \cps]),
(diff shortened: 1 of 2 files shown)


telephon (changes requested):
The variable `snd` holds a signal of arbitrary size (we know the size, it has `numChannels`). Your implementation will work for stereo only.

Also there is a whitespace indentation problem, maybe you need to check (tabs not spaces).

telephon:
Here is roughly how you could do it, but it's not yet right for the stereo case.

```supercollider
(
{
	var reverbPan = 0.5;
	var in = (0..7);
	var sig = SinOsc.ar(2 ** in * 100);
	var offset = reverbPan * in.size;
	var all = PanAz.ar(in.size, sig, (0..7) + offset * 2 / in.size, width: 1, orientation: 0);
	all.sum
}.plot
)
```

telephon:
@f-f do you want to work further on it?

f-f:
@telephon sorry I went on to other things and did not manage to work further on this. I'll close it

PR #127, opened 2019-04-25
Fix superprimes and superwavemechanics so pan works

bgold-cosmos:


telephon (commented):
are you sure you want to use a stereo mix to then pan to numChannels? Also, the `scramble` only happens when the synthdef is built, that is when you completely start supercollider. Maybe there is a better solution round the corner?

telephon:
@bgold-cosmos any follow up? It would be good to either merge or close soon.

bgold-cosmos:
Whoops, yeah, got sidetracked thinking about multichannel expansions.  I'm traveling the next few days but I'll put together something soon.

bgold-cosmos:
I think this is much better although at the moment I don't have the setup to thoroughly test multichannel output.

library/default-synths-extra.scd
```diff
@@ -480,13 +481,13 @@ SynthDef(\superwavemechanics, {|out, pan, freq, sustain, accelerate, detune=0, v
 	var env, sound, i, r, d, bwr, sz;
 	freq = freq * XLine.ar(1, exp(accelerate), sustain);
 	env = EnvGen.ar(Env.sine(1.0), timeScale:sustain, doneAction:2);
-	i = (0..7).scramble;
+	i = Array.fill(8, {TIRand.kr(0, 3, Impulse.kr(8))});
```
telephon: why trigger at 8 Hz? Also, if `i` is used in line 487 in a `Rand` it will never be retriggered.

library/default-synths-extra.scd
```diff
@@ -465,7 +465,8 @@ SynthDef(\superprimes, {|out, pan, freq, sustain, accelerate, speed=1, detune=0,
 	sound = SinOsc.ar(freq * (1..9)*[1+(detune/12),1-(detune/12)] / 2.0, phase)
     	* LPF.ar(Decay.ar(trig, (primes+3)*sustain/10), voice.linexp(0,2,30,18000));
 	sound = GVerb.ar(sound.softclip, 10*voice.linexp(0,5,1,0.01), 10).tanh;
-	OffsetOut.ar(out, DirtPan.ar(0.5*Splay.ar(sound.scramble.flat), ~dirt.numChannels, pan, env));
+	sound = SplayAz.ar(9, sound.flat, center:LFNoise2.kr(sustain).range(0,8));
```
telephon: Is there a particular reason why 9 channels?

telephon (changes requested):

library/default-synths-extra.scd
```diff
@@ -481,10 +481,10 @@ SynthDef(\superwavemechanics, {|out, pan, freq, sustain, accelerate, detune=0, v
 	var env, sound, i, r, d, bwr, sz;
 	freq = freq * XLine.ar(1, exp(accelerate), sustain);
 	env = EnvGen.ar(Env.sine(1.0), timeScale:sustain, doneAction:2);
-	i = Array.fill(8, {TIRand.kr(0, 3, Impulse.kr(8))});
+	i = Array.fill(8, {TIRand.kr(0, 3, Impulse.kr(sustain/8))});
```
telephon: because the 8 `Impulse`Ugens are all the same, you can just have one of them outside the Array and use it as an input for all.

`i = TIRand.kr(0, 3, Impulse.kr(sustain/8).dup(8)))` will expand and do the same.

library/default-synths-extra.scd
```diff
@@ -481,10 +481,10 @@ SynthDef(\superwavemechanics, {|out, pan, freq, sustain, accelerate, detune=0, v
...
 	r = [3,5,8,9]/8;
 	detune = min(detune, 1.5);
-	d = Rand(1.6*i - detune, 1.7*i + detune).round(0.25);
+	d = TRand.ar(1.6*i - detune, 1.7*i + detune, Impulse.ar(0)).round(0.25);
```
telephon: because you are triggering only once (`Impulse.ar(0)`), you can use a `Rand` instead:

`d = Rand.ar(1.6 * i - detune, 1.7 * i + detune).round(0.25);`

telephon (changes requested):
yes better!

I've added a few hints for (minor) improvement...

telephon (approved):

PR #128, opened 2019-04-25
Fix synths speed to rate

bgold-cosmos:
Use `rate` instead of `speed` for softsynths that were interpreting `speed` in some non-standard way.


telephon (commented):
what is the usual way of using `rate` from within tidal? I just saw that my own implementations in default-synths are not quite consistent.

bgold-cosmos:
`rate` is a new keyword, so the idea is to separate out whatever these synths are doing from sample playback rate

I took a look at default-synths but it seemed to me that the synths that actually made use of the `speed` parameter were using it for a playback rate (from a recorded sample buffer).  Since that seems appropriate, I left them alone.

telephon:
I had thought that it might be a good idea to be able to reuse the cutup patterns in tidal also for synths. This is why I sometimes have used `rate = (begin + 1) * (speed + Sweep.kr(1, accelerate));`, which could now be: `rate = (begin + 1) * (speed + Sweep.kr(1, accelerate)) * rate;`.

Then the `rate` can be used in various places, wherever it makes sense.

But in general, the question is whether we want to make analogies between synths and samples in terms of frequency.

bgold-cosmos:
From the feedback at lurk.org it seems that a common use case might be something like `s "{sample sample supersynth}%2" # speed "1.2 1.5"`, where people are using the speed for sample pitch control but really don't want it doing something weird to the synth.  So I tried to look through the synths and remove `speed` unless it was doing something directly analogous to playback rate (and pitch) modification.  In your synths, I think that's mostly the case, but I wrote a few that were using 'speed' to control an envelope rate or something, and it seems that was just confusing everyone.

telephon:
Probably the best thing -- for the future -- is to make versions of synths that are variants of the default ones and that mimic the behavior of samples.

PR #135, opened 2019-07-18
dirt: 12 orbits instead of just two

### LINKED ISSUE #134: Increase default number of orbits
I think 10 (number of fingers) or 12 (most divisible low number) would be better than the current default number of orbits, which I think is 2, but my guess at where the default is set is here, and it seems to be 1:
 https://github.com/musikinformatik/SuperDirt/blob/dd378ea88b014b3e20dbe4bf77138c0eadbe06ce/classes/SuperDirt.sc#L40

I'm not sure whether we should change any default memory requirements etc to match.

When this is done I'll change the tidal default boot to send different channels to the different orbits via:
```
    d1 = p 1 . (|< orbit 0) 
    d2 = p 2 . (|< orbit 1) 
    d3 = p 3 . (|< orbit 2) 
```
I think new users will find this behaviour less surprising when it comes to using global effects.

telephon:
This fixes  #134.

### DIFF
--- a/superdirt_startup.scd
+++ b/superdirt_startup.scd
@@ -8,7 +8,7 @@ This is an example startup file. You can load it from your startup file
 // configure the sound server: here you could add hardware specific options
 // see http://doc.sccode.org/Classes/ServerOptions.html
 s.options.numBuffers = 1024 * 256; // increase this if you need to load more samples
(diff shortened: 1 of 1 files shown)


yaxu:
Great!

PR #138, opened 2019-07-29
Make delay conscious of cps

rseymour:
This is a major bug for me, might be a major breaker for folks w/ sessions using delay

https://github.com/tidalcycles/Tidal/issues/542

### DIFF
--- a/synths/core-synths-global.scd
+++ b/synths/core-synths-global.scd
@@ -75,7 +75,7 @@ CORE SYNTHDEFS FOR DIRT
 
 	if(\SwitchDelay.asClass.notNil) {
 
-		SynthDef("dirt_delay" ++ numChannels, { |dryBus, effectBus, gate = 1, delaytime, delayfeedback, delaySend = 1, delayAmp = 1, lock = 0, cps = 1|
+		SynthDef("dirt_delay" ++ numChannels, { |dryBus, effectBus, gate = 1, delaytime, delayfeedback, delaySend = 1, delayAmp = 1, lock = 0, cps|
 			var signal;
 			var input = In.ar(dryBus, numChannels);
 			var maxDelayTime = 4;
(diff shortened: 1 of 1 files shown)


rseymour:
actually upon messing with this more, I think I need to also replace that if() with just the reciprocal to get the desired effect.

telephon:
I don't understand why this fixes anything - this is curious... In `DirtOrbit`, the parameter `cps` is specified, so it should always be updated if the incoming message has it. You could check by adding in the SynthDef:

```
cps.poll;
```

Also it must be reciprocal, because a higher cps should result in a shorter delay.

rseymour:
yea I've kept it reciprocal and have the diff there being
```
-                       delaytime = delaytime * if(lock, reciprocal(cps), 1);
+                       delaytime = delaytime * reciprocal(cps);```

**Comment by rseymour** [comment]
yea you're correct the cps = 1 is being overridden correctly, it's the above fix (which maybe I can fix with the default for lock?) works for me.

**Comment by telephon** [comment]
One question: in your code that doesn't work:
```haskell
d1 $ "cp" # delay 0.9 # delayt 0.144
```
there is no `# lock 1`

Which is what you want if you want to beat lock.

Alternatively, you can set the default in sclang:

```supercollider
~dirt.set(\lock, 1)
```

Had you tried this?

rseymour:
Yea I had always figured tempo sync (aka lock) would be the default and none of the tidal docs refer to lock in the context of anything let alone delay.  This was user error, but also a documentation fail.  I can work on the docs.  lock = 1 should work and this PR can be deleted, but I would personally make that the default as I tend to know how long a cycle is from what I'm hearing, not in terms of seconds.

telephon:
Depends on the genre, and also the delay time I suppose!

rseymour:
editing the wiki now.

yaxu:
Is this now solved, i.e. should we close the issues too?

PR #145, opened 2019-09-15
Added looper.scd

f00b455:
This looper allows to record and playback in Tidal

### DIFF
--- a/hacks/looper.scd
+++ b/hacks/looper.scd
@@ -56,9 +56,9 @@ SynthDef(\fooRecord, { |bufnum|
 functions[\fooloopRec] = {
 	if(recording) { functions[\fooloopAdd].value };
 	counter = counter + 1;
-	if(counter == numBuffers, {counter = 0});
+	if(counter > recBufs.lastIndex, {counter = 0});
 	~server.makeBundle(lat, {
-		recSynth = Synth(\fooRecord, [bufnum: (recBufs @@ counter)], ~server);
+		recSynth = Synth(\fooRecord, [bufnum: (recBufs @ counter)], ~server);
 		recording = true;
 	})
 };
@@ -82,9 +82,8 @@ functions[\fooloopReset] = {
 	counter = 0;
 	recording = false;
 };
-// use the "diversion" key as a hook for playing the synth
-~dirt.orbits[0,0,0,0,0,0,0,0].defaultParentEvent[\diversion] = {
-	var div = functions[~s];
-	if(div.notNil) { div.value; 1.0 }
-};
+
+// add these functions to the dirt soundLibrary
+functions.keysValuesDo{ arg key, func ; ~dirt.soundLibrary.addSynth( key, (play: func)) };
+
 )
--- a/hacks/looper.scd
+++ b/hacks/looper.scd
@@ -1,0 +1,89 @@
+/*
+fooloop is inspired by chunk-recorder.scd
+This looper records from SoundIn and writes/replaces buffers directly to the SuperDirt soundLibrary.
+
+Copyright (C) 2019  Michael Hauck (f00b455)
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


hacks/looper.scd
```diff
@@ -0,0 +1,89 @@
...
+SynthDef(\fooRecord, { |bufnum|
+	var in = SoundIn.ar([0,1]);
+	RecordBuf.ar(in, bufnum, loop:0, doneAction:2);
+}).add;
+
+// start recording round robin to buffers
+functions[\fooloopRec] = {
+	if(recording) { functions[\fooloopAdd].value };
+	counter = counter + 1;
+	if(counter == numBuffers, {counter = 0});
```
telephon: `counter` is used as an index into `recBufs`. But the last possible index in `recBufs` is `numBuffers - 1`. You can either skip this lile completely, because you are using `@@` below (which is `wrapAt`), and the counter can just go up forever. Or you can check for  e.g.`counter >= recBufs.lastIndex` and then just use `@` or `at` below.

hacks/looper.scd
```diff
@@ -0,0 +1,89 @@
...
+functions[\fooloopReset] = {
+	"reset".postln;
+	~recBufs.do(_.free);
+	~recBufs = recBufs = { Buffer.alloc(~dirt.server, ~dirt.server.sampleRate * maxTime, ~dirt.numChannels) } ! numBuffers;
+	for(0,numBuffers-1, { arg i; ~dirt.soundLibrary.addBuffer(bufnamePrefix++i, (recBufs @@ i), false )});
+	counter = 0;
+	recording = false;
+};
+// use the "diversion" key as a hook for playing the synth
+~dirt.orbits[0].defaultParentEvent[\diversion] = {
```
telephon: there is a different way now add functions to superdirt without using the diversion.

You can simply write (e.g.):
`~dirt.soundLibrary.addSynth(\fooloopReset, (play: { … your function … }))`.

then you can event pass parameter names, via ~foo, ~bar.

telephon (changes requested):

hacks/looper.scd
f00b455: thanks, this is much more readable!

hacks/looper.scd
f00b455: thanks, i will fix it.

f00b455: `counter` is used in `fooloopAdd` to generate the name for the soundLibrary. So it has to be resetted. I changed the `@@` to `@` and updated the condition for the reset.

hacks/looper.scd
telephon: good! one little request, could you check again and adjust the formatting to make it more consistent? This is just because when people start copying and reusing the code, this matters (otherwise, it would be fine!).

hacks/looper.scd
```diff
@@ -0,0 +1,89 @@
...
+SynthDef(\fooRecord, { |bufnum|
+	var in = SoundIn.ar(soundIn) * EnvGen.ar(Env.linen(0.003, 23, 0.003, 1, \sine), doneAction: 0);
+	RecordBuf.ar(in, bufnum, loop:0, doneAction:2);
+}).add;
+
+// start recording round robin to buffers
+functions[\fooloopRec] = {
+	if(recording) { functions[\fooloopAdd].value };
+	counter = counter + 1;
+	if(counter > recBufs.lastIndex, {counter = 0});
```
telephon: use same syntax as above, ok?

hacks/looper.scd
```diff
@@ -0,0 +1,89 @@
...
+	"reset".postln;
+	~recBufs.do(_.free);
+	~recBufs = recBufs = { Buffer.alloc(~dirt.server, ~dirt.server.sampleRate * maxTime, soundIn.size) } ! numBuffers;
+	for(0,numBuffers-1, { arg i; ~dirt.soundLibrary.addBuffer(bufnamePrefix++i, (recBufs @@ i), false )});
+	counter = 0;
+	recording = false;
+};
+
+// add these functions to the dirt soundLibrary
+functions.keysValuesDo{ arg key, func ; ~dirt.soundLibrary.addSynth( key, (play: func)) };
```
telephon: either use all `{ arg key ...` or `{ |key| ... }` (I prefer the latter, but as long as it is consistent, it's fine!).

telephon (changes requested):

telephon (approved):
thanks°!

PR #149, opened 2019-10-10
time stretching of samples

bgold-cosmos:
This implements time-scaling for sample playback using an overlap-add (OLA) algorithm.

I tweaked the code with speed and flexibility as priorities rather than output sound quality, but for modest scaling amounts (0.8 to 1.3 or so) it usually sounds fairly good.  I'm not aware of any method that will consistently timescale transient-only sounds (most drums) in a "nice" way.

The timescaling _should_ play well with `speed`, `begin`, `end` and similar parameters, but I'm sure I haven't tested every possible combination.

Examples of usage in Tidal would be
```
timescale = pF "timescale"

d1 $ chop 4 $ s "rave" # timescale "[1 4] 2" # cps 0.5

d1 $ s "metal*4" # cps 1
 # up ("c g6 c c6" + "<c ef g c>" - 12)
 # timescale "1 3 1 2"
```

### DIFF
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -102,7 +102,7 @@ DirtEvent {
 			if(~legato.notNil) {
 				~delta * ~legato.value
 			} {
-				unitDuration = (unitDuration ? ~delta) * (~stretch ? 1);
+				unitDuration = (unitDuration ? ~delta) * (~timescale ? 1);
 				loop !? { unitDuration = unitDuration * loop.abs };
 			}
 		};
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -102,7 +102,7 @@ DirtEvent {
 			if(~legato.notNil) {
 				~delta * ~legato.value
 			} {
-				unitDuration = unitDuration ? ~delta;
+				unitDuration = (unitDuration ? ~delta) * (~stretch ? 1);
 				loop !? { unitDuration = unitDuration * loop.abs };
 			}
 		};
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -102,7 +102,8 @@ DirtEvent {
 			if(~legato.notNil) {
 				~delta * ~legato.value
 			} {
-				unitDuration = (unitDuration ? ~delta) * (~timescale ? 1);
+				unitDuration = unitDuration ? ~delta;
+				if (~timescale.notNil) {unitDuration = unitDuration * ~timescale };
 				loop !? { unitDuration = unitDuration * loop.abs };
 			}
 		};
--- a/hacks/looper.scd
+++ b/hacks/looper.scd
@@ -28,8 +28,9 @@ d1 $ silence
 -- play recorded loop 3 (bufnamePrefix++counter)
 d5 $ sound "fooloop3" # gain "0.7"
 
--- reset buffers.
-once $ sound "fooloopReset"
+-- reset buffers. But do only one time (d1 $ silence),
+-- since d1 $ once $ sound "fooloopReset" does not work at my installation (1.4.3)
+d1 $ sound "fooloopReset"
 
 */
 (
--- a/synths/core-modules.scd
+++ b/synths/core-modules.scd
@@ -63,7 +63,6 @@ this may be refacored later.
 						end: ~end,
 						loop: ~loop,
 						pan: ~pan,
-						timescale: ~timescale,
 						out: ~out
 					])
 				}
--- a/synths/core-modules.scd
+++ b/synths/core-modules.scd
@@ -37,20 +37,36 @@ this may be refacored later.
 	{ |dirtEvent|
 		if(~diversion.value.isNil) {
 			if(~buffer.notNil) {
-				// argumets could be omitted using getMsgFunc, but for making it easier to understand, we write them out
-				dirtEvent.sendSynth(~instrument,  [
-					bufnum: ~buffer,
-					sustain: ~sustain,
-					speed: ~speed,
-					freq: ~freq,
-					endSpeed: ~endSpeed,
-					begin: ~begin,
-					end: ~end,
-					loop: ~loop,
-					pan: ~pan,
-					timescale: ~timescale,
-					out: ~out
-				])
+				if(~timescale.notNil) {
+					dirtEvent.sendSynth(~stretchInstrument, [
+						bufnum: ~buffer,
+						sustain: ~sustain,
+						speed: ~speed,
+						freq: ~freq,
+						endSpeed: ~endSpeed,
+						begin: ~begin,
+						end: ~end,
+						loop: ~loop,
+						pan: ~pan,
+						timescale: ~timescale,
+						out: ~out
+					])
+				} {
+					// argumets could be omitted using getMsgFunc, but for making it easier to understand, we write them out
+					dirtEvent.sendSynth(~instrument,  [
+						bufnum: ~buffer,
+						sustain: ~sustain,
+						speed: ~speed,
+						freq: ~freq,
+						endSpeed: ~endSpeed,
+						begin: ~begin,
+						end: ~end,
+						loop: ~loop,
+						pan: ~pan,
+						timescale: ~timescale,
+						out: ~out
+					])
+				}
 			} {
 				if(~instrument.isNil) {
 					"module 'sound': instrument not found: %".format(~sound).postln
--- a/synths/core-modules.scd
+++ b/synths/core-modules.scd
@@ -48,7 +48,7 @@ this may be refacored later.
 					end: ~end,
 					loop: ~loop,
 					pan: ~pan,
-					stretch: ~stretch,
+					timescale: ~timescale,
 					out: ~out
 				])
 			} {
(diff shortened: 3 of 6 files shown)


telephon:
just to be sure: have you checked the PitchShift UGen for comparison? Your implementation has the advantage that it could be combined with the https://github.com/musikinformatik/WavesetsEvent Quark, that sounds much better to my ear than overlap add.

In the end, of course we shouldn't load the standard player with extra things, but add a switch when timestretch is set, to save electricity.

bgold-cosmos:
Whenever I've tried the PitchShift UGen I've gotten really choppy results regardless of windowSize.  However, I should check again to do a sound comparison, and I'll take a look at the Wavesets stuff.

I wasn't sure how best to conditionally apply the effect.  My understanding is it's not really possible to have true branching in a SynthDef, so would there have to be a whole new set of `dirt_sample_timescale_%_%` SynthDefs that would get called when the event's `~timeScale` is non-nil?  I tried to keep the extra computation to a minimum - all it really is doing is making a pair of phase indices for `BufRd` rather than a single one, and doing a little extra modulation on them so that they're not quite simple LFSaws.  I haven't rigorously compared CPU usage but it didn't seem to add much.

telephon:
Why don't you implement it as an effect?

bgold-cosmos:
The current version can compress time too, so to do that as an effect would require looking into the future.  I don't know how to do that except for samples that've already been loaded.  Also, I think working with the sample indexing will be much more efficient, it only ever has two overlapping windows --- whereas as an effect I think it would need to be implemented using a network of staggered delays, some potentially with a very long delay time for large stretches and long samples.  Might be interesting to try both, though...

bgold-cosmos:
The current version can compress time too, so to do that as an effect would require looking into the future.  I don't know how to do that except for samples that've already been loaded.  Also, I think working with the sample indexing will be much more efficient, it only ever has two overlapping windows --- whereas as an effect I think it would need to be implemented using a network of staggered delays, some potentially with a very long delay time for large stretches and long samples.  Might be interesting to try both, though...

bgold-cosmos:
OK, the timescaling has now been moved to a separate SynthDef so no extra work is done if it isn't used.

synths/core-modules.scd
```diff
@@ -37,19 +37,36 @@ this may be refacored later.
...
+						bufnum: ~buffer,
+						sustain: ~sustain,
+						speed: ~speed,
+						freq: ~freq,
+						endSpeed: ~endSpeed,
+						begin: ~begin,
+						end: ~end,
+						loop: ~loop,
+						pan: ~pan,
+						timescale: ~timescale,
```
telephon: we don't need timescale in this case, right?

synths/core-synths.scd
```diff
@@ -52,6 +50,47 @@ live coding them requires that you have your SuperDirt instance in an environmen
...
+			// playback speed
+			rate = Line.kr(speed, endSpeed, sustain) * (freq / 60.midicps);
+
+			// sample phase
+			// BufSampleRate adjusts the rate if the sound file doesn't have the same rate as the soundcard
+			//phase =  Sweep.ar(1, rate * BufSampleRate.ir(bufnum)) + (BufFrames.ir(bufnum) * begin);
+
+			numFrames = BufFrames.ir(bufnum);
+			windowSize = numFrames * timescale / 29.0;
+			windowSize = windowSize.clip(1000.0*timescale, 4410.0);
```
telephon: why not `1000 * timescale, 4410` ? Mixing floats and ints is perfectly fine. Unless you are using specific operators like `rand`, `div`, where it makes a (reasonable) difference.

synths/core-synths.scd
```diff
@@ -52,6 +50,47 @@ live coding them requires that you have your SuperDirt instance in an environmen
...
+			timescaleStep = windowSize / timescale;
+			sawrate = rate * BufSampleRate.ir(bufnum) / (absdif(begin, end) * timescale * numFrames);
+			phase = (speed.sign * LFSaw.ar(sawrate, 1)).range(begin,end) * numFrames * timescale;
+			index = (phase.div(timescaleStep) - [1.0, 0.0]) * (timescaleStep / timescale - timescaleStep) + phase;
+			windowIndex = phase - (phase.div(timescaleStep) - [1.0, 0.0] * timescaleStep);
+			window = exp(-0.5 * (windowIndex/windowSize - 0.5/0.1)**2);
```
telephon: this can be written as:
`exp((5 * windowIndex/windowSize - 2.5) ** 2)`

or maybe better as:
`exp(squared((5 * windowIndex/windowSize - 2.5)))`

why these numbers and not others?

hacks/looper.scd
```diff
@@ -28,8 +28,9 @@ d1 $ silence
 -- play recorded loop 3 (bufnamePrefix++counter)
 d5 $ sound "fooloop3" # gain "0.7"
 
--- reset buffers.
-once $ sound "fooloopReset"
+-- reset buffers. But do only one time (d1 $ silence),
+-- since d1 $ once $ sound "fooloopReset" does not work at my installation (1.4.3)
+d1 $ sound "fooloopReset"
```
telephon, line 33: Maybe `my installation` is ambiguous here?

classes/DirtEvent.sc
```diff
@@ -102,7 +102,7 @@ DirtEvent {
 			if(~legato.notNil) {
 				~delta * ~legato.value
 			} {
-				unitDuration = unitDuration ? ~delta;
+				unitDuration = (unitDuration ? ~delta) * (~timescale ? 1);
```
telephon: it would be clearer and more efficient to write:
```
unitDuration = (unitDuration ? ~delta);
if(~timescale.notNil) { unitDuration = unitDuration * ~timescale };
```

telephon (changes requested):
The extra complications introduced are worth it, I think! Thank you.

synths/core-synths.scd
bgold-cosmos: It's a gaussian centered at the middle of the window with a width (sigma) of 0.1 the window size.  The 0.1, though, is really just trial and error to find what sounds good for a wide variety of sounds.  But can definitely combine this into fewer operations.

hacks/looper.scd
bgold-cosmos: I accidentally merged in something from the `develop` branch here, and I thought I undid it but apparently not.  I didn't make this change.

synths/core-synths.scd
telephon: maybe it would make sense as a parameter that you can pass, with a good default (in any case it makes sense to avoid unnecessary operations on a signal).

synths/core-synths.scd
```diff
@@ -68,13 +68,13 @@ live coding them requires that you have your SuperDirt instance in an environmen
 
 			numFrames = BufFrames.ir(bufnum);
 			windowSize = numFrames * timescale / 29.0;
-			windowSize = windowSize.clip(1000.0*timescale, 4410.0);
+			windowSize = windowSize.clip(1000*timescale, 4410);
```
telephon, line 71: Will that work on all sample rates? If not, replace `4410` by `0.1 * SampleRate.ir`).

telephon:
btw. sorry, I had overlooked your changes.

synths/core-modules.scd
```diff
@@ -37,19 +37,35 @@ this may be refacored later.
...
-					freq: ~freq,
-					endSpeed: ~endSpeed,
-					begin: ~begin,
-					end: ~end,
-					loop: ~loop,
-					pan: ~pan,
-					out: ~out
-				])
+				if(~timescale.notNil) {
+					dirtEvent.sendSynth(~stretchInstrument, [
```
telephon, line 41: I'd change this to:

```
var instrument, args;
...
args = [ ... ]; // the shorter arg list
if(~timescale.notNil) { 
	instrument = ~stretchInstrument;
	args = args.add(\timescale).add(~timescale)
} { 
	instrument = ~instrument 
};

dirtEvent.sendSynth(instrument, args);
…
```
Then you don't need the same list twice.

synths/core-modules.scd
```diff
@@ -37,19 +37,35 @@ this may be refacored later.
...
+						endSpeed: ~endSpeed,
+						begin: ~begin,
+						end: ~end,
+						loop: ~loop,
+						pan: ~pan,
+						timescale: ~timescale,
+						out: ~out
+					])
+				} {
+					// argumets could be omitted using getMsgFunc, but for making it easier to understand, we write them out
```
telephon, line 55: typo: argumets –> arguments

telephon (changes requested):

yaxu:
Shall we make these smallish changes and @bgold-cosmos ?

bgold-cosmos:
I'll refresh my memory on this and look into getting it updated.  A lot of things got put on pause during 2020.  I suspect given all the SuperDirt changes lately it'll be much cleaner to just redo this as a new PR - the semester is just starting Monday here but if I don't do something within a week or so don't feel bad about pinging me again.

telephon:
I think it won't be hard to resolve any conflicts.

bgold-cosmos:
I've redone this, see #229

Issue #115, opened 2019-02-21
Add renick's SOS drum recipe synth defs ?


madskjeldgaard:
Hello, just an idea but how about adding Renick Bell's really great supercollider interpretations of the classic sound on sound drum recipes ? Copypasted below from sccode.org. [sccode link](
https://doc.sccode.org/1-5aD/fork)

They need a bit of work to make them conform to the SuperDirt style but they sound great imo. 

```supercollider


SynthDef(\SOSkick,
	{ arg out = 0, freq = 50, mod_freq = 5, mod_index = 5, sustain = 0.4, amp = 0.8, beater_noise_level = 0.025;
	var pitch_contour, drum_osc, drum_lpf, drum_env;
	var beater_source, beater_hpf, beater_lpf, lpf_cutoff_contour, beater_env;
	var kick_mix;
	pitch_contour = Line.kr(freq*2, freq, 0.02);
	drum_osc = PMOsc.ar(	pitch_contour,
				mod_freq,
				mod_index/1.3,
				mul: 1,
				add: 0);
	drum_lpf = LPF.ar(in: drum_osc, freq: 1000, mul: 1, add: 0);
	drum_env = drum_lpf * EnvGen.ar(Env.perc(0.005, sustain), 1.0, doneAction: 2);
	beater_source = WhiteNoise.ar(beater_noise_level);
	beater_hpf = HPF.ar(in: beater_source, freq: 500, mul: 1, add: 0);
	lpf_cutoff_contour = Line.kr(6000, 500, 0.03);
	beater_lpf = LPF.ar(in: beater_hpf, freq: lpf_cutoff_contour, mul: 1, add: 0);
	beater_env = beater_lpf * EnvGen.ar(Env.perc, 1.0, doneAction: 2);
	kick_mix = Mix.new([drum_env, beater_env]) * 2 * amp;
	Out.ar(out, [kick_mix, kick_mix])
	}

//DrumSynths SC Example - SOS Drums by Renick Bell, renick_at_gmail.com
// recipes from Gordon Reid in his Sound on Sound articles
// SOSkick -------
// http://www.soundonsound.com/sos/jan02/articles/synthsecrets0102.asp
// increase mod_freq and mod_index for interesting electronic percussion
	).add;


SynthDef(\SOSsnare,
	{arg out = 0, sustain = 0.1, drum_mode_level = 0.25,
	snare_level = 40, snare_tightness = 1000,
	freq = 405, amp = 0.8;
[...]

PR #150, opened 2019-10-15
Add Renick Bell's SOS synths

yacodes:
An attempt to configure @renickbell's SOS drums for SuperDirt usage. Includes:
* **soskick**
* **soshats**
* **sossnare**
* **sostoms**

synthesizers, based on https://github.com/musikinformatik/SuperDirt/issues/115 and https://github.com/brunoruviaro/SCLOrkSynths

### DIFF
--- a/library/default-synths-extra.scd
+++ b/library/default-synths-extra.scd
@@ -523,3 +523,145 @@ SynthDef(\superreese,  {|out, pan, freq, sustain, accelerate, detune=0, voice=0|
 	OffsetOut.ar(out, DirtPan.ar(sound, ~dirt.numChannels, pan, env));
 }).add;
 );
+
+(
+// DrumSynths SC Example - SOS Drums by Renick Bell, renick@gmail.com
+// recipes from Gordon Reid in his Sound on Sound articles
+// SOSkick — http:www.soundonsound.com/sos/jan02/articles/synthsecrets0102.asp
+// increase pitch1 and voice for interesting electronic percussion
+// Adapted to SuperDirt by Aleksandr Yakunichev, hi@ya.codes
+//
+// midinote – controls the root note of the kick
+// pitch1 - controls modulation frequency in Hz (min: 0, max: SampleRate.ir / 2)
+// voice - controls modulation input phase in radians (min: 0, max: your sanity)
+// pitch2 - controls WhiteNoise amplitude (min: 0, max: 1)
+// speed - controls WhiteNoise sweep (min: 0, max: 1)
+SynthDef(\soskick, {
+  arg out, pan, freq = 65, pitch1 = 0, speed = 0.3, voice = 1, sustain = 1, pitch2 = 0.00;
+  var beater, source, env;
+
+  env = Env.perc(0.005, sustain).ar(Done.freeSelf);
+
+  source = Line.ar(freq * 2, freq, 0.02);
+  source = PMOsc.ar(source, pitch1, voice);
+  source = LPF.ar(source, 1000);
+
+  beater = WhiteNoise.ar(pitch2.clip);
+  beater = HPF.ar(beater, 500);
+  beater = LPF.ar(beater, XLine.ar(6000, 500, speed.clip * 0.1));
+  beater = beater * Env.perc.ar;
+
+  OffsetOut.ar(out, DirtPan.ar((source + beater) * 3.dbamp, ~dirt.numChannels, pan, env));
+},
+metadata: (
+  credit: "Renick Bell",
+  category: \drums,
+  tags: [\pitched, \bass, \sos]
+)).add;
+);
+
+(
+// Retrieved from http://sccode.org/1-5aD
+// DrumSynths SC Example - SOS Drums by Renick Bell, renick@gmail.com
+// Recipes from Gordon Reid in his Sound on Sound articles
+// SOShats — http:www.soundonsound.com/sos/Jun02/articles/synthsecrets0602.asp
+// Modified by Bruno Ruviaro and Josh Mitchell 8/19
+// Adapted to SuperDirt by Aleksandr Yakunichev, hi@ya.codes
+//
+// resonance – bandpass filter resonance value (min: 0, max: 1)
+// pitch1 - oscillator modulation in radians (min: 0, max: SampleRate.ir / 2)
+SynthDef(\soshats, {
+  arg out = 0, pan = 0, freq = 220, pitch1 = 238.5, resonance = 1, sustain = 0.5;
+  var source, envelope, bpf, hpf;
+
+  source = Mix(PMOsc.ar(Pulse.ar(freq), freq * [1.34, 2.405, 3.09, 1.309], pitch1 * [1, 0.22, 0.014, 0.0038]));
+
+  bpf = BPF.ar(
+    source,
+    XLine.kr(15000, 9000, sustain),
+    Clip.ir(resonance, 0, 1),
+    Env.perc(0.005, sustain, curve: -4).ar
+  );
+
+  hpf = HPF.ar(
+    source,
+    XLine.kr(9000, 12000, sustain),
+    Env.perc(0.005, sustain, curve: -4).ar
+  );
+
+  envelope = Env.perc(0.005, sustain).ar(Done.freeSelf);
+
+  OffsetOut.ar(out, DirtPan.ar((bpf + hpf) * (-5).dbamp, ~dirt.numChannels, pan, envelope));
+},
+metadata: (
+  credit: "Renick Bell",
+  category: \drums,
+  tags: [\pitched, \hihats, \sos]
+)).add;
+);
+
+(
+// Retrieved from http:sccode.org/1-5aD
+// DrumSynths SC Example - SOS Drums by Renick Bell, renick_at_gmail.com
+// recipes from Gordon Reid in his Sound on Sound articles
+// SOStom — http:www.soundonsound.com/sos/Mar02/articles/synthsecrets0302.asp
+// Modified by Bruno Ruviaro and Josh Mitchell 8/19.
+// Adapted to SuperDirt by Aleksandr Yakunichev, hi@ya.codes
+//
+// voice - controls modulation input phase in radians (min: 0, max: your sanity)
+SynthDef(\sostoms, {
+  arg out, pan, sustain = 0.5, freq = 261.626, voice = 0.5;
+  var envelope, source;
+
+  source = PMOsc.ar(Saw.ar(freq * 0.9), freq * 0.85, voice, mul: 6.dbamp);
+  source = source + SinOsc.ar([freq, freq * 0.8]);
+  source = Mix.ar(source);
+  source = LeakDC.ar(source + Crackle.ar(2)) * (-3).dbamp;
+
+  envelope = Env.perc(0.005, sustain, (-4).dbamp, -6).ar(Done.freeSelf);
+
+  OffsetOut.ar(out, DirtPan.ar(source, ~dirt.numChannels, pan, envelope));
+},
+metadata: (
+  credit: "Renick Bell",
+  category: \drums,
+  tags: [\pitched, \tom, \sos]
+)).add;
+);
+
+(
+// Retrieved from http:sccode.org/1-5aD
+// DrumSynths SC Example — SOS Drums by Renick Bell, renick_at_gmail.com
+// recipes from Gordon Reid in his Sound on Sound articles
+// SOSsnare — http:www.soundonsound.com/sos/Mar02/articles/synthsecrets0302.asp
+// Latch.ar(WhiteNoise.ar(0.1), Impulse.ar(nyquist * 2)) is added aliasing for effect
+// Modified by Bruno Ruviaro and Josh Mitchell 8/19.
+// Adapted to SuperDirt by Aleksandr Yakunichev, hi@ya.codes
+//
+// voice - controls modulation input phase in radians (min: 0, max: your sanity)
+// semitone - modulation frequency in semitones of fundamental
+// pitch1 - resonance filter frequency (Hz)
+// resonance - resonance of bandpass and resonz filters (min: 0, max: 1)
+SynthDef(\sossnare, {
+  arg out, pan, freq = 405, voice = 0.385, semitone = 0.452, pitch1 = 2000, resonance = 0.1, sustain = 0.5;
+  var source, noise, envelope;
+
+  source = PMOsc.ar(Saw.ar(freq * 0.85), freq * semitone, voice);
+
+  noise = Latch.ar(WhiteNoise.ar(0.1), Impulse.ar(1700 * 2));
+  noise = BRF.ar(noise, 4 * pitch1, resonance, 0.5);
+  noise = BRF.ar(noise, 2.5 * pitch1, resonance, 0.5);
+  noise = BRF.ar(noise, 1.8 * pitch1, resonance, 0.5);
+  noise = BRF.ar(noise, pitch1, resonance, Env.perc(0.005, sustain, -4).ar);
+  noise = Resonz.ar(noise, pitch1, 1, 40);
+
+  envelope = Env.perc(0.005, sustain, 0.5, -4).ar(Done.freeSelf);
+
+  OffsetOut.ar(out, DirtPan.ar((source + noise) * 4.dbamp, ~dirt.numChannels, pan, envelope));
+},
+metadata: (
+  credit: "Renick Bell",
+  category: \drums,
+  tags: [\pitched, \snare, \sos]
+)).add;
+);


library/default-synths-extra.scd
```diff
@@ -523,3 +523,157 @@ SynthDef(\superreese,  {|out, pan, freq, sustain, accelerate, detune=0, voice=0|
...
+
+  beaterSource = WhiteNoise.ar(beaterNoiseLevel);
+  beaterHpf = HPF.ar(in: beaterSource, freq: 500, mul: 1, add: 0);
+
+  lpfCutoffContour = Line.kr(6000, 500, 0.03);
+
+  beaterLpf = LPF.ar(in: beaterHpf, freq: lpfCutoffContour, mul: 1, add: 0);
+  beaterEnv = beaterLpf * EnvGen.kr(Env.perc, 1.0);
+
+  kickMix = Mix.new([drumEnv, beaterEnv]) * 2;
```
telephon: you can just add `drumEnv + beaterEnv`

library/default-synths-extra.scd
```diff
@@ -523,3 +523,157 @@ SynthDef(\superreese,  {|out, pan, freq, sustain, accelerate, detune=0, voice=0|
...
+  // Initial envelopes
+  initAtt = 0.005, initAmp = 1, initStart = 15000, initEnd = 9000,
+  // Body envelopes
+  bodyAtt = 0.005, bodyAmp = 1, bodyStart = 9000, bodyEnd = 12000;
+
+  var root, initialEnv, initialSweep, initial, bodyEnv, bodySweep, body, snd, env;
+
+  root = Pulse.ar(freq: freq, width: 0.5, mul: 1);
+  root = PMOsc.ar(carfreq: root, modfreq: freq * [1.34, 2.405, 3.09, 1.309], pmindex: rootIndex * [1, 0.22, 0.014, 0.0038]);
+  root = Mix.new(root);
```
telephon: you can write `Mix(root)`, sounds nicer to me.

library/default-synths-extra.scd
```diff
@@ -523,3 +523,157 @@ SynthDef(\superreese,  {|out, pan, freq, sustain, accelerate, detune=0, voice=0|
...
+
+  initialEnv = Env.perc(attackTime: initAtt, releaseTime: sustain / 2, curve: curve).kr;
+  initialSweep = Line.kr(start: initStart, end: initEnd, dur: sustain / 2);
+  initial = BPF.ar(in: root, freq: initialSweep, rq: rq, mul: initialEnv * initAmp);
+
+  bodyEnv = Env.perc(attackTime: bodyAtt, releaseTime: sustain / 2, curve: curve / 2).kr;
+  bodySweep = Line.kr(start: bodyStart, end: bodyEnd, dur: sustain / 2);
+  body = HPF.ar(in: root, freq: bodySweep, mul: bodyEnv * bodyAmp);
+
+  snd = Mix.ar([initial, body]) * amp;
```
telephon: same here, just add. (see below for more)

telephon (changes requested):

library/default-synths-extra.scd
```diff
@@ -523,3 +523,157 @@ SynthDef(\superreese,  {|out, pan, freq, sustain, accelerate, detune=0, voice=0|
...
+(
+// Retrieved from http://sccode.org/1-5aD
+// DrumSynths SC Example - SOS Drums by Renick Bell, renick@gmail.com
+// Recipes from Gordon Reid in his Sound on Sound articles
+// SOShats — http:www.soundonsound.com/sos/Jun02/articles/synthsecrets0602.asp
+// Modified by Bruno Ruviaro and Josh Mitchell 8/19
+SynthDef(\soshats, {
+  arg
+  // Standard Values
+  out = 0, pan = 0, freq = 6000, amp = 0.8, curve = -4, rootIndex = 238.5, rq = 1, sustain = 0.1,
```
bgold-cosmos: For this and the other synths, I believe Tidal+SuperDirt will set certain parameter defaults before the SynthDef is even called.  For example, if in Tidal you use `d1 $ s "soshats"` then freq will be automatically set to 261.6 (C5), amp to 0.4, sustain to 1 (depending on the pattern and cps), and pan to 0.5.  So this synth is probably going to sound weird unless you do `d1 $ s "soshats/10" # n "fs9"` or something, which is maybe not the intent.

yacodes: Yeah, you're totally right. TidalCycles sets these parameters by default, but all the synths actually sound pretty nice with default TC's parameters too.

The intent to leave freq and amp was to make these parameters available from TC controls and to be them as predictable as possible. I don't really think that SuperDirt's synths must sound incredible with no parameters provided to them from the end-user, it's kinda tricky though.

Maybe you have some suggestions on how to resolve this issue?

bgold-cosmos: (I'm going to use `soshats` as the main example since it's the most extreme case, but these apply to the other drums too)

- amp

It's only 6 dB so I don't know how much it matters. It could either be left as is, or you could just multiply by the factor of two at the end of the SynthDef
- freq

One method would be to multiply any incoming freq by 20 (or 16 if you like powers of two) so that the default `n` results in something close to 6000 Hz. This has the advantage that the default sounds close to as intended but the user can still mess around with `n` to get different pitches.

If there's really no discernible pitch though, another option would be to rename the parameter entirely so that `n` doesn't affect it, hardcode in the 6000 Hz, and possibly allow rescaling using `pitch1` (already defined in Tidal but rarely used) or a similar parameter.   For example, use `6000 * pitch1` anywhere `freq` is currently used.

- sustain

Sustain is a bit tricky since SuperDirt goes to a lot of effort to automatically provide a sustain to "fit" the duration into the "slot" coming from Tidal, which is great for something like a piano, but not really how drums usually work.  Do you really want the hat to decay twice as fast if someone does `s "soshats*8"` instead of `s "soshats*4"`?  All that said, I don't know how noticeable it really is in the end - you're already using fixed attack times which will help things sound consistent.  It might be worth experimenting a little bit with really fast/slow sequences to see if you might want to wrap the release time in a `clip`, which would keep things from getting too weird at the extremes.

yacodes: @bgold-cosmos Thanks for your suggestions!

* amp
I've decided to drop out this parameter from synthdefs and completely rely on SuperDirt's env and gain controls.

* freq
I do think that frequency in drums makes a big difference and the main issue with multiplying its value by powers of two or any other number is kinda confusing when you're using `note "e3"` in TC, for example. The user's intention is to hear exactly `e3` note, but in the case of multiplication, he will hear a whole different pitch. 

* sustain
I think it's the same with `freq` param. Why should we stop the users from getting freaky? 

---

Actually, thanks a lot for reminding me about `pitch1` and also `voice` parameters, which we can use for synth's modifications. I will prepare all the synths in a while.

yacodes:
@telephon @bgold-cosmos I feel like the `d634728` commit should be fine — can you check it out, please?

The synths' descriptions:
### soshats
(parameters: #n, #pitch1, #resonance, #sustain)
* `pitch1` — controls modulation index
* `resonance` — controls sweep resonance

### soskick
(parameters: #n, #pitch1, #voice, #sustain)
* `pitch1` — modulation frequency
* `voice` — modulation index

### sostoms
(parameters: #n, #sustain, #voice)
* `voice` — modulation index

### sossnare
(parameters: #n, #sustain, #voice, #semitone, #pitch1, #resonance)
* `voice` — modulation index
* `semitone` — modulation frequency deviation
* `pitch1` — filters frequency
* `resonance` — filters resonance

library/default-synths-extra.scd
telephon: re freq and sustain parameters: the difficulty is only to have good defaults, and good (that is interesting and somewhat transparent) deviations from defaults when something is specified. Sometimes you may want to interpret freq and sustain in a particular way from within a SynthDef.

Sometimes I write:

```supercollider
SynthDef(\x, { |freq = 440, speed = 1|
   var ratio = freq / 440;
   var trig = Impulse.ar(speed * ratio);
[…]
})
```

library/default-synths-extra.scd
```diff
@@ -578,7 +578,7 @@ SynthDef(\soshats, {
 
   initialEnv = Env.perc(attackTime: 0.005, releaseTime: sustain / 2, curve: -4).kr;
   initialSweep = Line.kr(start: 15000, end: 9000, dur: sustain / 2);
-  initial = BPF.ar(in: root, freq: initialSweep, rq: voice, mul: initialEnv);
+  initial = BPF.ar(in: root, freq: initialSweep, rq: resonance, mul: initialEnv);
```
telephon: could you please add a default?

library/default-synths-extra.scd
```diff
@@ -642,32 +637,26 @@ metadata: (
...
+  drumModeEnv = Env.perc(attackTime: 0.005, releaseTime: 0.1, level: 0.5, curve: curve).kr(doneAction: Done.freeSelf);
   drumMode = SinOsc.ar(freq: freq * 0.53, mul: drumModeEnv);
   drumMode = drumMode + SinOsc.ar(freq: freq, mul: drumModeEnv);
-  drumMode = drumMode + PMOsc.ar(carfreq: Saw.ar(freq * 0.85), modfreq: freq * modHarmonic, pmindex: timbreIndex, mul: drumModeEnv * 10);
-  drumMode = Mix.ar(drumMode) * drumModeAmp;
+  drumMode = drumMode + PMOsc.ar(carfreq: Saw.ar(freq * 0.85), modfreq: freq * semitone, pmindex: voice, mul: drumModeEnv * 10);
+  drumMode = drumMode * 0.25;
 
-  snareEnv = Env.perc(attackTime: att, releaseTime: rel, curve: curve).kr;
+  snareEnv = Env.perc(attackTime: 0.005, releaseTime: 0.1, curve: curve).kr;
```
telephon: better `...).ar` for such short attack times will sound better I think.

telephon (commented):
just a minor thing, sorry for the late reply.

library/default-synths-extra.scd
```diff
@@ -523,3 +523,146 @@ SynthDef(\superreese,  {|out, pan, freq, sustain, accelerate, detune=0, voice=0|
...
 }).add;
 );
+
+(
+// DrumSynths SC Example - SOS Drums by Renick Bell, renick@gmail.com
+// recipes from Gordon Reid in his Sound on Sound articles
+// SOSkick — http:www.soundonsound.com/sos/jan02/articles/synthsecrets0102.asp
+// increase pitch1 and voice for interesting electronic percussion
+SynthDef(\soskick, {
+  arg out, pan, freq = 50, pitch1 = 5, voice = 5, sustain = 1, nl = 0.025;
```
telephon: usually, we call filter frequency `ffreq`. Would that be a good name?

library/default-synths-extra.scd
yacodes: what do you mean by `default`? `resonance` value is `1` by default, described at the `arg` section in `soshats` synthdef

library/default-synths-extra.scd
yacodes: Yeah, I think it will be fine, but then you cannot control frequency from `TidalCycles` (except `pF "ffreq" 200`, but it is quite unobvious).

What do you think about mapping frequency, passed from TidalCycles to only one octave for the kick? I mean this:
`d1 $ s "soskick" # freq 63` (or `c3`) -> stays unchanged and used "as is" 63Hz
`d1 $ s "soskick" # freq 130` (or `c4`) -> maps to 63Hz as well, cause it is out of defined limits

And the limits is the lower octave 63-123Hz

How about that behaviour? This Synthdef will always sound fine + you can tune it with TidalCycles as well!

yacodes:
@telephon Thanks for the review! I couldn't help myself and have made a couple of rewrites for more clean code and some conceptual aspects of the code, considering your comments of course. Now I am extremely excited about the sound and functionality of the resulting synths – it will be incredible if you can review the code one more time, please!

library/default-synths-extra.scd
```diff
@@ -523,3 +523,148 @@ SynthDef(\superreese,  {|out, pan, freq, sustain, accelerate, detune=0, voice=0|
...
+//            to avoid unpleasant sounds
+// pitch1 - controls modulation frequency in Hz (min: 0, max: SampleRate.ir / 2)
+// voice - controls modulation input phase in radians (min: 0, max: your sanity)
+// pitch2 - controls WhiteNoise amplitude (min: 0, max: 1)
+// speed - controls WhiteNoise sweep (min: 0, max: 1)
+SynthDef(\soskick, {
+  arg out, pan, midinote = 24, pitch1 = 0, speed = 0.3, voice = 1, sustain = 1, pitch2 = 0.00;
+  var beater, source, freq, env;
+
+  freq = (Wrap.ir(midinote, 24, 36)).midicps;
```
telephon: Using midinote as an argument won't allow you to use `# freq` in tidal. Perhaps it would be better if you used `freq` as an argument directly, and then limit that by
`freq = freq.clip(24.midicps, 36.midicps)`
For this, I used clip, but of course if you have reasons for using wrap, that would also be reasonable ;-)

library/default-synths-extra.scd
```diff
@@ -523,3 +523,148 @@ SynthDef(\superreese,  {|out, pan, freq, sustain, accelerate, detune=0, voice=0|
...
+  env = Env.perc(0.005, sustain).kr(Done.freeSelf);
+
+  source = Line.kr(freq * 2, freq, 0.02);
+  source = PMOsc.ar(source, pitch1, voice);
+  source = LPF.ar(source, 1000);
+
+  beater = WhiteNoise.ar(Clip.kr(pitch2));
+  beater = HPF.ar(beater, 500);
+  beater = LPF.ar(beater, XLine.ar(6000, 500, Clip.ir(speed, 0, 1) * 0.1));
+  beater = beater * Env.perc.kr;
```
telephon: will sound better as `Env.perc.ar`.

library/default-synths-extra.scd
```diff
@@ -523,3 +523,148 @@ SynthDef(\superreese,  {|out, pan, freq, sustain, accelerate, detune=0, voice=0|
...
+// voice - controls modulation input phase in radians (min: 0, max: your sanity)
+// pitch2 - controls WhiteNoise amplitude (min: 0, max: 1)
+// speed - controls WhiteNoise sweep (min: 0, max: 1)
+SynthDef(\soskick, {
+  arg out, pan, midinote = 24, pitch1 = 0, speed = 0.3, voice = 1, sustain = 1, pitch2 = 0.00;
+  var beater, source, freq, env;
+
+  freq = (Wrap.ir(midinote, 24, 36)).midicps;
+  env = Env.perc(0.005, sustain).kr(Done.freeSelf);
+
```
telephon, line 544: will sound better as `Env.perc.ar`.

library/default-synths-extra.scd
```diff
@@ -523,3 +523,148 @@ SynthDef(\superreese,  {|out, pan, freq, sustain, accelerate, detune=0, voice=0|
...
+  source = Line.kr(freq * 2, freq, 0.02);
+  source = PMOsc.ar(source, pitch1, voice);
+  source = LPF.ar(source, 1000);
+
+  beater = WhiteNoise.ar(Clip.kr(pitch2));
+  beater = HPF.ar(beater, 500);
+  beater = LPF.ar(beater, XLine.ar(6000, 500, Clip.ir(speed, 0, 1) * 0.1));
```
telephon: normally, we write `speed.clip(0, 1)` but not an issue.

telephon (changes requested):
just a few minor things. Thanks!

library/default-synths-extra.scd
telephon: (maybe. But as it is noise, maybe no worries …)

library/default-synths-extra.scd
yacodes: Sure, didn't know about `freq` parameter in Tidal, thanks!

I'm using `Wrap.ir` for nice usage in Tidal: `d1 $ s "soskick" # n "e"`. Where you can just plug any note without octave specification in `n` param and it will adopt kick to it.

yacodes:
@telephon Some more changes according to your comments! :))

library/default-synths-extra.scd
telephon: ah I see, very good. So the best would be to define a parameter `root = 24`, which is the reference midinote, and then write `freq = freq.wrap(root.midicps, (root + 12).midicps)`

telephon: … then you could work in different harmonic reference frames.

renickbell:
Hi.

Rather than limit the freq of the kick, I'd recommend just suggesting
values in documentation. It's been really so long since I wrote that code,
but I sort if remember using the kick as an alternate tom at the time.

I also endorse the comment above about letting users get freaky if they
want.

Best,

Renick

telephon:
Yes, @renickbell, perhaps the functionality of making one instrument having the same note as another modulo some value would be better implemented in tidal than hard-wired in the SynthDef.

yacodes:
@telephon `Wrap.ir` is removed!

PR #153, opened 2019-12-02
Six-op FM synth

loopier:
@yaxu suggested in [this forum thread](https://forum.toplap.org/t/array-parameters-for-sc-custom-synth/942/17?u=loopier) to add the synth to the repo.

This is my first ever pull-request and contribution to another repo.  Please forgive me if I'm not doing things properly.   I'm open to suggestions and guidance on doing so.


bgold-cosmos:
This is fun!  The output always seems to be at peak level, it might be worth turning it down a bit; I think somewhere around -10dB would put it about even with a typical Tidal sample. 

Since there are so many parameters, it might be nice to have some examples or "presets".  This could be in SuperDirt or even on the Tidal end, e.g. `fmpreset1 = mod11 1 # detune1 1 # detune2 1 # mod12 20 # ratio1 1.02 # ratio2 2`

loopier:
I will review the output level, thanks.

Where should I add the examples or "presets" in Tidal?
If I do it in SuperDirt, should I do it as new SynthDefs, or is there a better way to do it?

bgold-cosmos:
I don't know about the "best" way to do it, but one example might be to look what I did with the 'voice' parameter in superhammond.  Basically a big multidimensional array of parameter values that you can dial "presets" from using `SelectX` with the `voice` parameter.

loopier:
I reduced the output level by -15dB and added a few random presets that can be accessed with the `voice` parameter as you suggested.  

I also added a `feedback` parameter.  With manual feedback mods (mod11, mod22, ... mod66) it must be set to `1` in order to hear them.

I thought it would  be more fun to have random presets than predefined ones, so the musician won't know how they are going to sound.  The presets are generated with random values every time the SynthDef is evaluated.  They will be different on every machine, or even on every session if the SynthDefs are evaluated on startup. Are they?

library/default-synths-extra.scd
```diff
@@ -523,3 +523,109 @@ SynthDef(\superreese,  {|out, pan, freq, sustain, accelerate, detune=0, voice=0|
...
+// ratio2 2.3  (op2 frequency ratio)
+// mod11 0.5   (op1 feedback)
+// mod12 0.78  (op1 modulation amount by op2)
+// detune1 0.2 (op1 detune)
+// eglevel12   (op1 EG level2)
+// egrate11    (op1 EG rate1) -- higher values go FASTER!
+
+(
+SynthDef(\superfm, {
+	var sustain = \sustain.kr(1);
```
telephon, line 557: if your controls are controlled one-per-event it would be much more efficient to use `.ir` instead of `.kr`.

loopier: I'm not sure how events are treated in SuperDirt.  Is an event triggered every time a Tidal message is recieved?  How do `wave` arguments in Tidal translate to SuperDirt?  I tried to follow the other SynthDefs structure, and since `rate` is not specified their arguments default to `.kr`, don't they?

telephon: >I'm not sure how events are treated in SuperDirt. Is an event triggered every time a Tidal message is recieved?

Yes

> How do `wave` arguments in Tidal translate to SuperDirt?

they are just passed on without translation, and received by the synth if it defines the argument. In principle, you could also send lists/arrays, but this is not yet implemented in the tidal-sclang communication.

> I tried to follow the other SynthDefs structure, and since rate is not specified their arguments default to .kr, don't they?

the SynthDefs have a list of rates specified in the end (`[\ir, \ir, \ir]` e.g.), this is how this is normally done. But you can conveniently specify this in your SynthDef, where you define arguments in the style `symbol.kr(defaultValue)`, or `symbol.ir(defaultValue)`.

loopier: sure.  I was thinking it might be good to change some of them to `\ir` but then realized that it's nice to be able to modulate all the parameters.  What are your thoughts?

telephon: Yes, you could leave it as it is, maybe you can then map controls in supercollider to continuous processes. Only if you want to have the most efficient implementation for tidal, then you'd use initial rate, because it saves electricity.

telephon:
@loopier ready to merge or do you want to still change something?

loopier:
> @loopier ready to merge or do you want to still change something?

I'm ready!  Just made a last revision and fixed the conflict.  Thanks!

loopier:
Thanks!

telephon:
thank you!

telephon:
I just noticed that the random numbers in the presets are fixed in each synthdef, so they will be the same for each note, but different on each restart of superdirt. Is this intended? 

In principle, it would be possible to have presets in superdirt, not in the synthdef. But that's for another time, I think.

loopier:
> I just noticed that the random numbers in the presets are fixed in each synthdef, so they will be the same for each note, but different on each restart of superdirt. Is this intended?
> 
Yes, it is.  I thought it would be more fun to have random presets than predefined ones, so the musician won't know how they are going to sound.

> In principle, it would be possible to have presets in superdirt, not in the synthdef. But that's for another time, I think.
Sure!

telephon:
OK, good!

telephon:
btw. the detunes are integers between -7 and 7, is that intended? `Array.fill(6, {rand2(7)}),`

loopier:
yep, also intended.

loopier:
But I may change it to floats if you reckon it would be better

telephon:
No, not necessarily. It is an aesthetic choice. With whole number detunes, you get a more predictable pattern of multiples of one second beatings, I suppose? I haven't tried.

Issue #154, opened 2019-12-03
Support NRPN?


yaxu:
NRPN is a way to send high resolution data over MIDI CC messages
 https://www.elektronauts.com/t/nrpn-tutorial-how-to/46669

For example to change the x parameter on this alesis micron You send first send a two-byte controller identifier for it using CCs 99 and 98 `99:0 98:44` (these being cc number, cc value pairs). You then send a two-byte value over CCs 44 (msb) and 38 (lsb), e.g. `44:6 38:4 = 128 * 6 + 4 = a value of 772.

The id and value is always sent over those four cc numbers. You might send the id once followed by a series of values in quick succession.

Here's an example of someone sending a value using supercollider:
  https://sccode.org/1-57O

I tried this in tidal:
  `once $ cc "[99:0,98:44,6:0,38:9]" # s "midi"`

Unfortunately the list comes through in a random order, so the sequence isn't set properly. This works:
  `once $ cc "[99:0 98:44 6:7 38:0] ~@10000" # s "midi"`

But it would be great if it could be sent as a name:value pair with supercollider doing the work to break it into bytes for sending as CC:s. e.g.:
```
once $ nrpn "44*16" # val sine # s "midi"`
```

yaxu:
I've had a go, but just can't fathom supercollider out at all. Despite a lot of moving things around, I just get lots of `ERROR: Message 'asInteger' not understood.`

telephon:
Maybe some of the variables are nil? The `MIDIOut` tries to interpret the nils as integers, and fails (probably the thrower is posted further below in the error message).

telephon:
Are you sure your `midicmd` is really nil? The whole initialisation of the variabls happens only if this is the case (line 44).

yaxu:
Yes midicmd is nil. I think maybe the `~` variables are global and so aren't held in the function closure?

yaxu:
Yes that was it. Will test next time I'm with the synth

yaxu:
I redid my changes, seems to work well now.

yaxu:
This isn't working with negative values.

yaxu:
Any ideas what this should be, to support sending negative numbers?
```
					var nrpnMSB, nrpnLSB, valMSB, valLSB;
					~val = ~val ? 0;
					nrpnLSB = ~nrpn % 128;
					nrpnMSB = (~nrpn - nrpnLSB) / 128;
					valLSB  = ~val % 128;
[...]

PR #155, opened 2019-12-04
basic support of NRPN, ref #154

yaxu:


classes/DirtEventTypes.sc
```diff
@@ -38,18 +38,18 @@ DirtEventTypes {
...
 					if(~ccn.notNil) { midicmd = \control; ~ctlNum = ~ccn };
 					if(~ccv.notNil) { midicmd = \control; ~control = ~ccv };
 					if(~nrpn.notNil) {
 						midicmd = \control;
-						~nrpnLSB = ~nrpn % 128;
-                        ~nrpnMSB = (~nrpn - ~nrpnLSB) / 128;
-						~valLSB = ~val % 128;
-                        ~valMSB = (~val - ~valLSB) / 128;
+						nrpnLSB = ~nrpn % 128;
+                        nrpnMSB = (~nrpn - nrpnLSB) / 128;
```
telephon: could you check the formatting? Mabe you need to use be tabs, not spaces.

classes/DirtEventTypes.sc
```diff
@@ -81,24 +80,23 @@ DirtEventTypes {
...
-							midiout.control(~chan, 99, ~nrpnMSB);
-							midiout.control(~chan, 98, ~nrpnLSB);
-							midiout.control(~chan, 6,  ~valMSB);
-							midiout.control(~chan, 38, ~valLSB)
+						if (nrpnMSB.notNil) {
+							midiout.control(chan, 99, nrpnMSB);
+							midiout.control(chan, 98, nrpnLSB);
+							midiout.control(chan, 6,  valMSB);
+							midiout.control(chan, 38, valLSB)
 						}
```
telephon: usually, we use the syntax
```
if(something) {

} {

}
```

classes/DirtEventTypes.sc
```diff
@@ -81,24 +80,23 @@ DirtEventTypes {
...
 					{
-						if (~nrpn.notNil) {
+						if (nrpnMSB.notNil) {
 							thisThread.clock.sched(latency, {
-								midiout.control(0,99,0);
-								midiout.control(~chan, 99, ~nrpnMSB);
-								midiout.control(~chan, 98, ~nrpnLSB);
-								midiout.control(~chan, 6,  ~valMSB);
-								midiout.control(~chan, 38, ~valLSB)
+								midiout.control(chan, 99, nrpnMSB);
```
telephon: you don't need to duplicate this code. Jus define a function further up and call it in both cases. Functions are first-class, so they are just something like:

```
var sendNRPN = {
	midiout.control(chan, 99, nrpnMSB);
	midiout.control(~chan, 98, ~nrpnLSB);
	midiout.control(chan, 98, nrpnLSB);
	midiout.control(~chan, 6,  ~valMSB);
	midiout.control(chan, 6,  valMSB);
	midiout.control(~chan, 38, ~valLSB)
	midiout.control(chan, 38, valLSB)
};

```

telephon (changes requested):

classes/DirtEventTypes.sc
yaxu: Is it any better now?

classes/DirtEventTypes.sc
yaxu: Thanks, done!

classes/DirtEventTypes.sc
telephon: no, not yet! Not sure what it is. You can select and hit tab in the IDE for autoformat. And show invisible characters to double check that there are tabs not spaces.

yaxu: Ok sorry, sorted now. ("Show spaces and tabs" seems to have no effect on my 3.10.3 supercollider compiled under linux.)

telephon (approved):

PR #159, opened 2019-12-17
support infering multiple midi commands from parameters, will fix #157

### LINKED ISSUE #157: Multiple midi commands in one message
There is logic to infer midi command from the values in a message. I think this could be used to allow more than one midi command in one message.

yaxu:
Not fully working yet - a noteOn is always triggered

### DIFF
--- a/classes/DirtEventTypes.sc
+++ b/classes/DirtEventTypes.sc
@@ -41,7 +41,7 @@ DirtEventTypes {
 				var freq, lag, sustain;
 				var args, midiout, hasGate, midicmd, latency, chan;
 				var sendNRPN, schedmidi, schedmidicmd;
-				var hasNote;
+				var hasNote = ~n != \none;
 
 				midiout = ~midiout.value;
 
@@ -53,7 +53,6 @@ DirtEventTypes {
 				midicmd = ~midicmd;
 				~chan = ~midichan ? 0;
 				chan = ~chan;
-				hasNote = (not (~midinote.isNil && ~freq.isNil && ~n.isNil && ~octave.isNil));
 
 				lag = ~lag + (~latency ? 0);
 				latency = lag; // for now.
--- a/classes/DirtEventTypes.sc
+++ b/classes/DirtEventTypes.sc
@@ -36,6 +36,7 @@ DirtEventTypes {
 
 
 	midiEvent = (
+			midiOutNotFoundError: DirtPartTimeError(2, "midi device is nil."),
 			play: #{
 
 				var freq, lag, sustain;
@@ -46,10 +47,8 @@ DirtEventTypes {
 				midiout = ~midiout.value;
 
 				if(midiout.isNil) {
-					"midi device is nil.".postln;
-					^true
+					~midiOutNotFoundError.throw;
 				};
-
 				midicmd = ~midicmd;
 				~chan = ~midichan ? 0;
 				chan = ~chan;
--- a/classes/DirtEventTypes.sc
+++ b/classes/DirtEventTypes.sc
@@ -47,87 +47,86 @@ DirtEventTypes {
 
 				if(midiout.isNil) {
 					"midi device is nil.".postln;
-					return(true)
-				};
 
-				midicmd = ~midicmd;
-				~chan = ~midichan ? 0;
-				chan = ~chan;
+				}
+				{
+					midicmd = ~midicmd;
+					~chan = ~midichan ? 0;
+					chan = ~chan;
 
-				lag = ~lag + (~latency ? 0);
-				latency = lag; // for now.
-				schedmidi = if (latency == 0.0) {
-					{|f| f.value}
-				} {
-					{|f| thisThread.clock.sched(latency, f)}
-				};
-				schedmidicmd = {
-					|cmd|
-					var func;
-					func = Event.default[\midiEventFunctions][cmd];
-					args = func.valueEnvir.asCollection;
-					schedmidi.value({midiout.performList(cmd, args)});
-					if (midicmd.notNil && midicmd === cmd) {
-						midicmd = nil
+					lag = ~lag + (~latency ? 0);
+					latency = lag; // for now.
+					schedmidi = if (latency == 0.0) {
+						{|f| f.value}
+					} {
+						{|f| thisThread.clock.sched(latency, f)}
 					};
-				};
-
-				// guess MIDI events from parameters
-				if(~ccn.notNil && ~ccv.notNil) {
-					var ctlNum = ~ccn; // TODO - also check for ~ctlNum ?
-					var control = ~ccv;
-					schedmidi.value({ midiout.control(chan, ctlNum, control) });
-				};
-
-				if(~nrpn.notNil) {
-					var nrpnMSB, nrpnLSB, valMSB, valLSB;
-					~val = ~val ? 0;
-					nrpnLSB = ~nrpn % 128;
-					nrpnMSB = (~nrpn - nrpnLSB) / 128;
-					valLSB  = ~val % 128;
-					valMSB  = (~val - valLSB) / 128;
-					schedmidi.value({
-						midiout.control(chan, 99, nrpnMSB);
-						midiout.control(chan, 98, nrpnLSB);
-						midiout.control(chan, 6,  valMSB);
-						midiout.control(chan, 38, valLSB)
-					});
-				};
-
-				if(~progNum.notNil)   { var num = ~progNum;   schedmidi.value({ midiout.program(chan, num) })};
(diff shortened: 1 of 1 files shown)


telephon (approved):

classes/DirtEventTypes.sc
```diff
@@ -38,86 +38,93 @@ DirtEventTypes {
...
-						};
-					};
-					if(~progNum.notNil) { midicmd = \program };
-					if(~polyTouch.notNil) { midicmd = \polyTouch };
-					if(~midibend.notNil) { midicmd = \bend; ~val = ~midibend; };
-					if(~miditouch.notNil) { midicmd = \touch; ~val = ~miditouch; };
-					if(midicmd.isNil) { midicmd = \noteOn }; // if still nil
+				if(midiout.isNil) {
+					"midi device is nil.".postln;
+					return(true)
```
telephon: hm? you probably mean 
```
^true
```

classes/DirtEventTypes.sc
```diff
@@ -38,86 +38,93 @@ DirtEventTypes {
...
-
-				func = Event.default[\midiEventFunctions][midicmd];
-				args = func.valueEnvir.asCollection;
 
-				latency = lag; // for now.
+				// guess MIDI events from parameters
+				if(~ccn.notNil && ~ccv.notNil) {
+					var ctlNum = ~ccn; // TODO - also check for ~ctlNum ?
+					var control = ~ccv;
+					schedmidi.value({ midiout.control(chan, ctlNum, control) });
```
telephon, line 77: you can replace `({` by `{`, like in the `if`statement. Just looks better.

classes/DirtEventTypes.sc
```diff
@@ -38,86 +38,93 @@ DirtEventTypes {
...
-
-				~ctlNum = ~ctlNum ? 0;
-				~chan = ~midichan ? 0;
-
-				func = Event.default[\midiEventFunctions][midicmd];
-				args = func.valueEnvir.asCollection;
 
-				latency = lag; // for now.
+				// guess MIDI events from parameters
+				if(~ccn.notNil && ~ccv.notNil) {
```
telephon: more efficient is:
```
~ccn.notNil and: { ~ccv.notNil }
```
That will be inlined

classes/DirtEventTypes.sc
```diff
@@ -38,86 +38,93 @@ DirtEventTypes {
...
-				if(~uid.notNil and: { midiout.notNil }) {
-					~uid = midiout.uid    // mainly for sysex cmd
+				latency = lag; // for now.
+				schedmidi = if (latency == 0.0) {
+					{|f| f.value}
+				} {
+					{|f| thisThread.clock.sched(latency, f)}
+				};
+				schedmidicmd = {
+					|cmd|
```
telephon: better move the args up `{ |cmd|`

telephon (changes requested):

classes/DirtEventTypes.sc
telephon: that is `schedmidi.value { midiout.control … `

classes/DirtEventTypes.sc
yaxu: I didn't want superdirt to keep looking for other event types. At this point I think right event type is found, there's just a problem with it. I haven't actually tested this control flow path, though..

yaxu: Tested now, and it seems `return` isn't a thing in sclang. I'll make a big if, as before.

bgold-cosmos: I think `^` is just SC syntax for `return`, right?

yaxu: Aha I thought it was `not`, fixed back to use `^true`

telephon: I'm sorry, I didn't read carefully enough.

In methods of objects, you can return in the middle, with the caret `^`. But the functions here are not called as methods, but are functions in an event. So indeed, you can't use `^` here.

There are two ways to deal with this in these cases:

1. a "big if"
2. the method `block` allows you to jump out by a continuation-like construct. It passes a function to your outer function as first argument that you can use as an escape door by calling it with the return value:

```supercollider

(
x = block { |return|
	
	inf.do { |i|
		if(0.4.coin) { return.value(i) };
	};
	
};
x.postln;
)
```
Maybe version (1) is easier in the present case.

yaxu: Hm, but if I change the code to this:
```
		if(midiout.isNil) {
					"midi device is nil.".postln;
					^true
				};
				"oh".postln;
```
Then it seems to work fine. At least, if I run `~dirt.soundLibrary.addMIDI(\midi2, nil);`, then I get `midi device is nil.`, but no `oh`. (The `oh` does appear as expected, when the midi device is not nil)

telephon: Interesting. I think it just seems fine, but isn't. I'm pretty sure that you are exiting into the wrong context.

classes/DirtEventTypes.sc
```diff
@@ -38,89 +38,93 @@ DirtEventTypes {
...
-						};
-					};
-					if(~progNum.notNil) { midicmd = \program };
-					if(~polyTouch.notNil) { midicmd = \polyTouch };
-					if(~midibend.notNil) { midicmd = \bend; ~val = ~midibend; };
-					if(~miditouch.notNil) { midicmd = \touch; ~val = ~miditouch; };
-					if(midicmd.isNil) { midicmd = \noteOn }; // if still nil
+				if(midiout.isNil) {
+					"midi device is nil.".postln;
+					^true
```
telephon: better not exit like this. Also `true` doesn't make sense as a return value, because in all the other cases, the event returns itself. So when you call:
```
x = event.play
```
you sometimes get an `Event` for x, sometimes `true`. Why not throw an error?

telephon: hm, on a second thought, you may not want to flood the post window with heaps of error messages. I'll think of an error that you can throw and that won't do that.

telephon (changes requested):

classes/DirtEventTypes.sc
telephon: Ok, done. Now you can add an error to the midievent that you can then throw.

```supercollider
midiEvent = (
			midiOutNotFoundError: DirtPartTimeError(2, "midi device is nil."),
			play: #{

			<…>
			if(midiout.isNil) {
					~midiOutNotFoundError.throw;
			}


```

yaxu: Thanks, I've pushed the change, but it's not working for me, `DirtPartTimeError` doesn't seem to exist:
```
Execution warning: Class 'DirtPartTimeError' not found
ERROR: Message 'new' not understood.
```

telephon (approved):
good! Have you tested if it works? (Btw. you can leave the semicolon away if it doesn't separate statements)

yaxu:
No it's not working just yet, see https://github.com/musikinformatik/SuperDirt/pull/159#discussion_r360314734

classes/DirtEventTypes.sc
telephon: sorry, I forgot to merge. Now it should work.

yaxu: Great, that's better. It generates a large call stack error dump, but otherwise works fine.

telephon: is it bad that it is a large dump? That's usually the way it is done for "proper" errors, i.e. those you can't ignore

yaxu: I find these stack dumps difficult to interpret and in this case the information isn't useful. I think it could be a problem to fill the postbuffer in a performance, if I mistyped the name of a midi device.

classes/DirtEventTypes.sc
```diff
@@ -36,91 +36,94 @@ DirtEventTypes {
...
+					{|f| f.value}
+				} {
+					{|f| thisThread.clock.sched(latency, f)}
+				};
+				schedmidicmd = { |cmd|
+					var func;
+					func = Event.default[\midiEventFunctions][cmd];
+					args = func.valueEnvir.asCollection;
+					schedmidi.value { midiout.performList(cmd, args) };
+					if (midicmd.notNil && midicmd === cmd) {
```
telephon, line 68: this should be: `midicmd.notNil && (midicmd === cmd)` sclang always binds to the left. More efficient in this case is: `midicmd.notNil and: { midicmd === cmd }` which inlines everything so you have fewer function calls.

yaxu: Ok thanks, yes this bug resulted in doubling up of events. I created a new PR for this.

PR #168, opened 2019-12-26
stop doubling of events inferred by parameters if midicmd is set

yaxu:


classes/DirtEventTypes.sc
```diff
@@ -92,9 +95,9 @@ DirtEventTypes {
 					});
 				};
 
-				if(~progNum.notNil)   { var num = ~progNum;   schedmidi.value({ midiout.program(chan, num) })};
-				if(~midibend.notNil)  { var val = ~midibend;  schedmidi.value({ midiout.bend(chan, val)    })};
-				if(~miditouch.notNil) { var val = ~miditouch; schedmidi.value({ midiout.touch(chan, val)   })};
+				if(~progNum.notNil)   { var num = ~progNum;   donecmd.value(\program); schedmidi.value({ midiout.program(chan, num) })};
```
yaxu, line 98: I'm curious why this works:
```
donecmd.value(\program); schedmidi.value({ midiout.program(chan, num) })};
```
But this doesn't:
```
schedmidi.value({ midiout.program(chan, num) })}; donecmd.value(\program);
```
The call of `donecmd` seemed to be delayed in the latter case?

telephon: `schedmidi` is delayed and happens after `donecmd` in both cases.

yaxu: I meant in the latter example, the command ends up being sent twice.

yaxu: Anyway it works as-is, I think this is ok for merging

telephon:
Although I think that now you can't send two commands in one event, right? Not sure what was intended. It is slightly messy in terms of functionality. Maybe you could add an explanation in a longer comment in the file?

yaxu:
@telephon Unless I've made a mistake, you can send multiple commands in one event, but only one of each type.

yaxu:
(The purpose of this bit of code was to make sure that a command is only sent once if its 'midicmd' is inferred from its parameters, as well as being set explicitly)

telephon:
Ok, I just don't get what `donecmd` is doing.

yaxu:
It unsets `midicmd` if its action has already been inferred and done.

telephon:
Ah I see. But wouldn't it have been easier to write instead of:
```supercollider
if(~progNum.notNil)   { var num = ~progNum;   donecmd.value(\program); schedmidi.value({ midiout.program(chan, num) })};
```

simply
```supercollider
if(~progNum.notNil or: { midicmd === \program })   { var num = ~progNum;  schedmidi.value({ midiout.program(chan, num) })};
```

yaxu:
But then there's a bit later on that deals with the midicmd variable in a generic way, so it needs to be unset.

PR #170, opened 2019-12-26
Topic/midi event improvements

telephon:
here are some minor improvements for clarity.

### DIFF
--- a/classes/DirtEventTypes.sc
+++ b/classes/DirtEventTypes.sc
@@ -43,6 +43,8 @@ DirtEventTypes {
 				var args, midiout, hasGate, midicmd, latency, chan;
 				var sendNRPN, schedmidi, schedmidicmd, donecmd;
 				var hasNote = ~n != \none, midiCommandPending = true;
+				var nrpnMSB, nrpnLSB, valMSB, valLSB;
+				var ctlNum, control, num, val, note;
 
 				midiout = ~midiout.value;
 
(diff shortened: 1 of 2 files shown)


classes/DirtEventTypes.sc
```diff
@@ -91,7 +92,7 @@ DirtEventTypes {
 						midiout.control(chan, 99, nrpnMSB);
 						midiout.control(chan, 98, nrpnLSB);
 						midiout.control(chan, 6,  valMSB);
-						midiout.control(chan, 38, valLSB)
+						midiout.control(chan, 38, valLSB);
```
yaxu, line 96: Does the trailing semicolon have a functional use here, or is it just convention?

hacks/sclang-dirt.scd
```diff
@@ -29,15 +29,20 @@ var clumps = names.asArray.scramble.curdle(0.1);
 Pdef(\x,
 	Pbind(
 		\type, \dirt,
-		\s, Pn(Plazy { var n = clumps.choose; Pfuncn({ n.choose }, 64) }),
+		\s, Pn(Plazy {
+			var n = clumps.choose;
+			postln(join(n, " ") ++ "\n");
```
yaxu, line 34: Is this postln meant to be here - looks liek a stray trace statement

yaxu (approved):

hacks/sclang-dirt.scd
telephon: I found it good to see which instruments are playing at a given moment (btw. sorry, the chenge is unrelated).

classes/DirtEventTypes.sc
telephon: no, I just found it looks better because it makes the series of commands look the same.

If there is a rule about the trailing semicolon, it is that it is good style to definitely leave it out where the return value of the function is used. In cases of multiple side effects it's fine to leave it. But all that isn't very strongly followed, so a bit of a matter of taste.

telephon:
One more thing, this is a little bit more important: in sclang preferences under `Interpreter`, please set `Post Inline Warnings` to true. The current versions will complain for all those who have set it to true.

It is good in terms of efficiency to move variable declarations outside of if/while statements. If this doesn't result in code that is well readable, functions can be defined and then called in the if statement.

yaxu:
I see, will do that, thanks

PR #173, opened 2020-01-16
Limit frequencies to filters

bgold-cosmos:


synths/core-synths.scd
```diff
@@ -124,14 +124,14 @@ live coding them requires that you have your SuperDirt instance in an environmen
 
 	SynthDef("dirt_hpf" ++ numChannels, { |out, hcutoff = 440, hresonance = 0|
 		var signal = In.ar(out, numChannels);
-		signal = RHPF.ar(signal, hcutoff.abs, hresonance.linexp(0, 1, 1, 0.001));
+		signal = RHPF.ar(signal, hcutoff.abs.clip(20,80e3), hresonance.linexp(0, 1, 1, 0.001));
```
telephon: just out of curiosity: how have you found that magic number?

bgold-cosmos: No magic, just late at night and I forgot about SampleRate.ir (I've updated the PR)

Some people are particular about high frequencies so I was reluctant to just chop at 20 kHz.  But such people would be using a high server sampling rate, so I think it's sensible to just filter at the Nyquist frequency.

telephon (approved):

telephon:
one more thing: could you do a rebase on develop? Just to get rid of all these unrelated merges.
`git rebase -i develop` then resolve any conflicts (I don't think there are any) and `git push -f` rewrite the history here.

bgold-cosmos:
rebase done, and FWIW I also needed a `git checkout --track upstream develop` first since my repo dated back before the name change from master.  Thanks for providing the missing step, I had been having trouble figuring out how to get things back in sync.

yaxu:
Thanks very much for this @bgold-cosmos, and thanks also for the git tip, @telephon

PR #184, opened 2020-03-12
Added class for interfacing with tidalcycles

Dsm0:
I though it would be nice to have a small set of functions to load sound libraries with, so I added a TidalInterface class and a small set of functions that can be used to load sound files from paths.


classes/TidalInterface.sc
```diff
@@ -0,0 +1,113 @@
+/*
+An attempt at giving Tidalcycles a direct interface to SuperDirt,
+so you could load samples, load synths, kill the server, etc. all from a Tidalcycles Ide
+
+Currently, the SuperDirt class intercepts osc messages containing parameters
+sent from Tidal, and sends them t
+*/
+
+InterfaceEvent {
```
telephon: Please rename: `DirtInterfaceEvent`

classes/TidalInterface.sc
```diff
@@ -0,0 +1,113 @@
+/*
+An attempt at giving Tidalcycles a direct interface to SuperDirt,
+so you could load samples, load synths, kill the server, etc. all from a Tidalcycles Ide
+
+Currently, the SuperDirt class intercepts osc messages containing parameters
```
telephon, line 5: this is not true, it doesn't intercept messages. You just need to send messages with a different command than `/play2`.

classes/TidalInterface.sc
```diff
@@ -0,0 +1,113 @@
...
+
+	var <event;
+	var soundLibrary;
+
+    *new { |event,soundLibrary|
+				// "in Interface".postln;
+        ^super.newCopyArgs(event,soundLibrary)
+    }
+
+	test {
```
telephon, line 20: better a more specific name.

classes/TidalInterface.sc
```diff
@@ -0,0 +1,113 @@
...
+		// server = ~server.value;
+
+		if(soundLibrary != nil)
```
telephon: you can use `.notNil`, that is more idiomatic.

classes/TidalInterface.sc
```diff
@@ -0,0 +1,113 @@
...
+			{\freeAllSoundFiles} {this.freeAllSoundFiles}
+			{\freeSoundFiles} {this.freeSoundFiles}
+			{\postSampleInfo} {this.postSampleInfo}
+			{\initFreqSynthWindow} {this.initFreqSynthWindow}
+	}
+
+	loadSynthDefs{
+		// "in loadSynthDefs".postln;
+		if(event[\filePath] != nil)
+			{soundLibrary.loadSynthDefs(event[\filePath].asString)}
```
telephon: check formatting

classes/TidalInterface.sc
```diff
@@ -0,0 +1,113 @@
...
+	loadSoundFile { 
+	// "in loadSoundFile".postln;
+		if(event[\filePath] != nil)
+			{soundLibrary.loadSoundFile(event[\filePath].asString)}
+			{"error: no path passed to loadSoundFile"};
+	}
+
+	freeAllSoundFiles {
+	// "in freeAllSoundFiles".postln;
+	soundLibrary.freeAllSoundFiles;
```
telephon, line 88: formatting. Also you can drop the semicolon here.

telephon (changes requested):
This is a great addition. Independent of my other review comments I think it would be better not to mix the sound event logic with the API here. You can just add another OSCFunc to the `netResponders` of the SuperDirt class. Then you can open different messages as an API. I hope this is clear, feel free to ask.

The idea is very welcome!

classes/SuperDirt.sc
```diff
@@ -223,26 +223,39 @@ SuperDirt {
...
 				index = event[\orbit] ? 0;
-				// event.postln;
-
-				//send shit to the InterfaceEvent class
 
 				if(warnOutOfOrbit and: { index >= orbits.size } or: { index < 0 }) {
 						"SuperDirt: event falls out of existing orbits, index (%)".format(index).warn
 				};
 
+<<<<<<< HEAD
```
telephon: there are still conflicts to resolve...

telephon (changes requested):
also: could you be so kind and separate the two pull requests (the change in the synths and the change in the OSC receiver?) Thanks!

telephon:
Please feel free to reopen if you have time to continue. Thanks!

Issue #186, opened 2020-03-24
midinote no longer works


kindohm:
With the latest superdirt 2f31f42be32113d0add2a6f8e573e7ee4ca2765d, I am no longer able to play MIDI notes with `midinote`:

`d1 $ midinote "42" # s "midi"`

There is no error displayed in the console, but the configured device does not play the note. I have not yet used a MIDI inspector to see if any MIDI data is being sent at all. 

Reverting to the previous commit I was using (e2d4619) resolved the issue.

yaxu:
I'm seeing the same.

`n` works, e.g. `once $ n "0" # sound "midi"`

However `note` doesn't work either.

yaxu:
Ok I'm happy with this now. The following is the behaviour I'm now seeing:

```
-- sends no midi (with no errors)
once $ s "midi"

-- sends note 64 on (and off)
once $ n 4 # s "midi"

-- also sends note 64 on (and off)
once $ note 4 # s "midi"

-- sends note 4 on (and off)
once $ midinote 4 # s "midi"

-- sends cc message (nonote)
once $ cc "30:30" # s "midi" 

-- sends cc message *and* midi note 64 on (and off)
once $ s "midi" # cc "30:30" # n 4

-- sends note 64 on/off (note wins)
once $ midinote 4 # note 4 # s "midi"

-- sends note 4 on/off (midinote wins)
once $ midinote 4 # n 4 # s "midi"

-- sends note 64 on/off (just once)
once $ n 4 # note 4 # s "midi"
```

PR #188, opened 2020-04-05
MIDI to recognise midinote and note parameters

yaxu:


classes/DirtOrbit.sc
```diff
@@ -186,6 +186,7 @@ DirtOrbit {
 			~cut = 0.0;
 			~unit = \r;
 			~n = \none; // sample number or note
+			~note = \none; // sample pitch or note
```
telephon: this would then break the automatic derivation of `note`from `n` if note is not specified (currently, `== nil`, so that `~midinote = #{ ~note ? ~n + (~octave * 12) };` will ues `n`).

classes/DirtEventTypes.sc
```diff
@@ -42,7 +42,8 @@ DirtEventTypes {
 				var freq, lag, sustain;
 				var args, midiout, hasGate, midicmd, latency, chan;
 				var sendNRPN, schedmidi, schedmidicmd, donecmd;
-				var hasNote = ~n != \none, midiCommandPending = ~midicmd.notNil;
+				var hasNote = not (~n == \none and: {~note == \none});
```
telephon: I don't understand why now `~n == \none`, while before it was `~n != \none` ...

classes/DirtOrbit.sc
yaxu: Ah yes.
Without this line, I'm having some problems working out whether note has been set. For instance this:
```
("Note is " + ~note).postln;
if (~note != \nil) {
  ("Note is not nil.").postln
};
```
produces:
```
Note is  nil
Note is not nil.
```
I'm missing something here I think!

classes/DirtEventTypes.sc
yaxu: I got a bit in knots with the nots and ended up sticking a 'not' in front.

classes/DirtOrbit.sc
telephon: try this:

```
("Note is " + ~note).postln;
if (~note != nil) {
  ("Note is not nil.").postln
};
```

or,

```
("Note is " + ~note).postln;
if (~note.notNil) {
  ("Note is not nil.").postln
};
```

yaxu: That worked, thanks!

classes/DirtEventTypes.sc
yaxu: Replaced with an or, possibly is clearer..

yaxu:
Should be ready to go now, ref https://github.com/musikinformatik/SuperDirt/issues/186

classes/DirtEventTypes.sc
```diff
@@ -42,7 +42,7 @@ DirtEventTypes {
 				var freq, lag, sustain;
 				var args, midiout, hasGate, midicmd, latency, chan;
 				var sendNRPN, schedmidi, schedmidicmd, donecmd;
-				var hasNote = not (~n == \none and: {~note == \none});
+				var hasNote = (~n != \none or: {~note != nil});
```
telephon: great! While you are at it, `notNil` is a bit faster, and more idiomatic.

you can check:
```supercollider
{ |x| x != nil }.def.dumpByteCodes
{ |x| x.notNil }.def.dumpByteCodes
```

so:

```supercollider
hasNote = (~n != \none) or: { ~note.notNil };
```

yaxu: Thanks, fixed that!

PR #197, opened 2020-05-04
panic button, ref https://github.com/tidalcycles/Tidal/issues/454, fixes #107

### LINKED ISSUE #107: Ctrl-dot
It'd be good to be able to do the equivalent of ctrl-dot from tidal, to silence running synths

Maybe by diverting the `ctrldot` sample name, e.g. via `once $ s "ctrldot"`

yaxu:


telephon (commented):
great!
It's not exactly the equivalent of cmd-./ctrl-. because it will only touch superdirt, nothing else. Maybe ```// all lights off! frees all running synths in superdirt```

yaxu:
Ok! But I tried it yesterday and superdirt stopped working, I think something about nodes not being found. I'll try to reproduce that.

telephon:
hm, this shouldn't happen. When it happens you could try to call `~dirt.orbits.do(_.initNodeTree)` and see if it works again then. If yes, then it might be a synchronisation problem. If no, there is something missing from the `initNodeTree` method.

telephon:
@alex have you tested it a bit more? In terms of code it looks ok.

yaxu:
I've had another go and it's worked well, so don't know what happened before.

```
panic = do hush
           once $ sound "superpanic"

d1 $ fast 32 $ sound "breaks152 bev"

panic
```
I think this is good to merge

PR #206, opened 2020-07-26
Fixed small pitch shifting bug in SynthDef and add a parameter

thgrund:
I have seen that there is a "dirt_pitch_shift" module in core_synths. Unfortunately it does not work as intended, because the pitch shift parameter "pitchRatio" in core_synths is called ``psratio`` and in core_modules ``psrate``. 
I fixed this typo (now it is called `psrate` everywhere) and was able to use it successfully on the tidal side.

I also added the timeDispersion parameter for PitchShift and named it `disp`.

### DIFF
--- a/synths/core-modules.scd
+++ b/synths/core-modules.scd
@@ -148,6 +148,7 @@ this may be refacored later.
 		dirtEvent.sendSynth('dirt_pitch_shift' ++ ~numChannels,
 			[
 				psrate: ~psrate,
+				disp: ~disp,
 				sustain: ~sustain,
 				out: ~out
 		])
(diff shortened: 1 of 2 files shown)


telephon (changes requested):
This looks very good, thank you! One request: a clearer name for `disp` would be `psdisp` – could you change this?

thgrund:
Sure, that's no problem. I renamed disp to psdisp. I like these clarification, because that way you can better associate psdisp with psrate.

telephon (approved):

telephon:
Thank you!

PR #213, opened 2020-10-23
dirt: allow to switch between several limiter types

yaxu:


synths/core-synths-global.scd
```diff
@@ -32,12 +32,13 @@ CORE SYNTHDEFS FOR DIRT
...
 
-		signal = Limiter.ar(signal);
-
-		// or if you like it harsher
-		//signal = tanh(signal);
-		// or:
-		//signal = distort(signal);
+		signal = Select.ar(limitertype,
+			[
+				signal,
```
telephon, line 37: maybe we can afford to add one: `softclip(signal * 0.5) * 2` ? This will be linear beween -1 and 1 and then start to distort.

telephon (approved):

synths/core-synths-global.scd
yaxu: That sounds good!
I was assuming that there was a single, global limiter, but this looks like it's per-sound? If so, I'd quite like a global one, so extra-loud things 'punch through' everything else.

telephon: No, the global synths are per orbit. There are ways to add a limiter to the output of the supercollider output channels. Would that be better, perhaps?

yaxu:
Shall we merge this for now?

Issue #131, opened 2019-06-09
Inconsistent sorting of files


yaxu:
It seems that given a folder with files `kicksnare.wav` and `kick.wav`, linux based systems sort the former first, and others sort the latter first. Probably because other systems ignore the extension when sorting. This means `d1 $ sound "gretsch:13"` gives a kick on some systems, and a kick-snare on others.

I made a quick script for finding problem folders:
```bash
#!/bin/bash
for d in */ ; do
  diff <(ls $d|tr ' ' '_'|xargs basename -s.wav -a|sort) <(ls $d|tr ' ' '_'|sort|xargs basename -s.wav -a) >/dev/null
  if [ $? -ne 0 ]; then
    echo "$d has a problem.."
  fi
done
```
There's quite a few:
```
auto/ birds/ east/ gretsch/ h/ if/ jungbass/ mash2/ mute/ peri/ popkick/ print/ speakspell/ speechless/ toys/ trump/ ul/
```

Not sure of the best approach to a fix - by renaming the files to make sorting unambiguous, by fixing superdirt to sort one way or the other, or maybe this is a bug in supercollider.

yaxu:
I've disambiguated the filenames for now with this

```perl
#!/usr/bin/perl -w

use strict;

my $dirname = shift @ARGV;

opendir(DIR, $dirname) or die "Could not open $dirname\n";

my @files;

while (my $filename = readdir(DIR)) {
  if ($filename =~ /(.*?)\.wav$/i) {
    push(@files,[$1,$filename]);
  }
}

@files = sort {$a->[0] cmp $b->[0]} @files;

my $n = 0;
foreach my $file (@files) {
  my $new = sprintf("%03d_$file->[1]", $n++);
  system("git", "mv", "$dirname/$file->[1]","$dirname/$new");
  print("rename($dirname/$file->[1], $dirname/$new)\n");
}

closedir(DIR);
```

bgold-cosmos:
Are you sure this is an issue for SuperCollider sort order and not just shell `sort`?  I think `sort` output can depend on the shell and environment variable settings, none of which likely matter to SuperCollider

yaxu:
I think it would be good for superdirt behaviour to be consistent across platforms, so I think it (or supercollider) needs to do its own sort. I'd vote on one that sorts based on the basename (without extension), as to me that seems least surprising.

telephon:
It'll be very easy to fix this. You just need to sort the paths after line 172:

[...]

PR #214, opened 2020-10-29
Sort sound library files by ascii code

ndr-brt:
As discussed in #131

### DIFF
--- a/classes/SuperDirt.sc
+++ b/classes/SuperDirt.sc
@@ -124,7 +124,7 @@ SuperDirt {
 		soundLibrary.loadOnly(names, path, appendToExisting )
 	}
 
-	loadSoundFileFolder { |folderPath, name, appendToExisting = false|
+	loadSoundFileFolder { |folderPath, name, appendToExisting = false, sortFiles = true|
 		soundLibrary.loadSoundFileFolder(folderPath, name, appendToExisting)
 	}
 
@@ -391,4 +391,3 @@ SuperDirt {
 
 
 }
-
--- a/classes/SuperDirt.sc
+++ b/classes/SuperDirt.sc
@@ -125,7 +125,7 @@ SuperDirt {
 	}
 
 	loadSoundFileFolder { |folderPath, name, appendToExisting = false, sortFiles = true|
-		soundLibrary.loadSoundFileFolder(folderPath, name, appendToExisting)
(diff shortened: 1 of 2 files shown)


telephon:
How do we handle breaking changes @yaxu ?

yaxu:
@telephon In tidal I release a major version (according to https://pvp.haskell.org/ ) and put some details in a CHANGELOG file

However with the discussion around ordering from metadata files, maybe it isn't worth the breaking change here?

yaxu:
I think this isn't a breaking change under linux btw. I think the difference between platforms is whether the extension is considered in the sorting.

telephon:
How should we proceed here? I think it is good to have a consistent ordering. 

On macOS there seems to be no difference, here is a test:
```supercollider

(
var k;
k = { |e| e[\bufferObject].path.basename };

~dirt.soundLibrary.bufferEvents.select { |x|
	x.order { |a, b| k.(a) < k.(b) } != (0..x.size-1)
};
)

```

telephon:
Good!  I had forgotten about your PR. Making it globally configurable seems like a bad idea in my version. Probably better as an argument to the function itself. Would you like to add this to yours?

```
loadSoundFileFolder { |folderPath, name, appendToExisting = false, sortFiles = true|

…
if(sortFiles) { files.sort };
…
```

ndr-brt:
@telephon done! added the argument also to the outer `loadSoundFileFolder`

classes/SuperDirt.sc
```diff
@@ -124,7 +124,7 @@ SuperDirt {
 		soundLibrary.loadOnly(names, path, appendToExisting )
 	}
 
-	loadSoundFileFolder { |folderPath, name, appendToExisting = false|
+	loadSoundFileFolder { |folderPath, name, appendToExisting = false, sortFiles = true|
 		soundLibrary.loadSoundFileFolder(folderPath, name, appendToExisting)
```
telephon: you have to pass the argument on to `soundLibrary.loadSoundFileFolder` …

telephon (changes requested):
just one little thing …

classes/SuperDirt.sc
ndr-brt: Yes, you're right, added it.

telephon (approved):

PR #220, opened 2021-01-23
dirt-synth: move gain and overgain to synth

telephon:
This will allow a continuous modulation of `amp` with bus mappings.

It is a bit more expensive, but I have tried to minimise by making `gain` and `overgain` unmodulatable.

### DIFF
--- a/synths/core-synths.scd
+++ b/synths/core-synths.scd
@@ -108,16 +108,17 @@ live coding them requires that you have your SuperDirt instance in an environmen
 	// the monitor does the mixing and zeroing of the busses for each sample grain
 	// so that they can all play in one bus
 
-	SynthDef("dirt_gate" ++ numChannels, { |out, in, sustain = 1, fadeInTime = 0.001, fadeTime = 0.001, amp = 1|
+	SynthDef("dirt_gate" ++ numChannels, { |out, in, sustain = 1, fadeInTime = 0.001, fadeTime = 0.001, amp = 1, gain = 1, overgain = 0|
 		var signal = In.ar(in, numChannels);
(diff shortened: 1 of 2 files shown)


telephon:
Note that this might change the behaviour in places where a synth has an explict `amp` argument and one uses `gain` on it. This is not currently the case in the synths in superdirt.

yaxu (approved):

yaxu:
Ah sorry I hadn't actually tested this out properly before release - it seems that `gain` and `overgain` aren't now working. `amp` is fine. I tried with samples and synths (supermandolin and superpiano). even `gain 0` has no effect.

synths/core-synths.scd
```diff
@@ -108,16 +108,17 @@ live coding them requires that you have your SuperDirt instance in an environmen
 	// the monitor does the mixing and zeroing of the busses for each sample grain
 	// so that they can all play in one bus
 
-	SynthDef("dirt_gate" ++ numChannels, { |out, in, sustain = 1, fadeInTime = 0.001, fadeTime = 0.001, amp = 1|
+	SynthDef("dirt_gate" ++ numChannels, { |out, in, sustain = 1, fadeInTime = 0.001, fadeTime = 0.001, amp = 1, gain = 1, overgain = 0|
 		var signal = In.ar(in, numChannels);
 		 //  doneAction: 14: free surrounding group and all nodes
 		var env = EnvGen.ar(Env([0, 1, 1, 0], [fadeInTime, sustain, fadeTime], \sin), doneAction: 14);
+		amp = amp * pow(gain.min(2) + overgain, 4);
```
telephon, line 115: Here ist should be ...

PR #227, opened 2021-01-26
dirt-event-types: make sure amp is evaluated before

### LINKED ISSUE #226: sclang: Fading a pattern with Pdef has issues
I'm running `scsynth` version 3.11.2 on Arch Linux and the `Quarks.gui` tells, the latest version of SuperDirt is installed.
I tried the [sclang-dirt.scd](https://github.com/musikinformatik/SuperDirt/blob/develop/hacks/sclang-dirt.scd) hack, but unfortunately I managed to crash something.

A minimal example to reproduce. Everything set up, so this plays the bass drum sample as expected:
```
(type:\dirt, orbit:0, s: \bd).play; 
```
Let's play a pattern:
```
(
Pdef(\x,
	Pbind(
		\type, \dirt,
		\s, \bd,
		\n, Pseq([1, 0, 0, 0], inf),  // four on the floor
		\dur, 0.5
	)
).play
)
```
If I set a `fadeTime` for the `Pdef` ...
```
Pdef(\x).fadeTime=8
```
...then it will crash as soon as a new pattern is evaluated:
```
(
Pdef(\x,
	Pbind(
		\type, \dirt,
		\s, \hh,  // hi-hats
		\dur, 0.25
	)
).play
)
```
The error is `ERROR: Message 'dbamp' not understood.` but the rest of the dump does not give me a clue what is going on. And `Pdef(\x)` will only play SuperDirt again, after I restarted both interpreter and server.
Can somebody please help?

telephon:
fixes: #226

### DIFF
--- a/classes/DirtEventTypes.sc
+++ b/classes/DirtEventTypes.sc
@@ -18,6 +18,7 @@ DirtEventTypes {
 				Error("dirt event: no dirt instance found.\n\n// You could try:\nSuperDirt.default = ~dirt;").throw;
 			};
 			~delta = ~delta ?? { ~stretch.value * ~dur.value };
+			~amp = ~amp.value;
 			~latency = ~latency ?? { dirt.server.latency };
 			if(~n.isArray) {
 				keys = currentEnvironment.keys.asArray;


telephon:
@yaxu can you merge these to all that matters?

yaxu:
Sure, thanks!

PR #210, opened 2020-10-05
Bus mapping

telephon:
Continuous control for dirt synths and a handshake for tidal.

Issue #196, opened 2020-05-02
long, 1-hour samples sound bad


kindohm:
When using very, _very_ long samples, SuperDirt seems to degrade the quality of the sample, as if the sample rate is downscaled. I haven't found the time cutoff point, but this effect can be produced with a 60-minute sample:

```
d1 $ s "extremelylongsample" # legato 1
```

The audible difference is very obvious and unpleasant.

telephon:
Yes, SuperDirt is not really made for playing long samples, the precision limit is rather around 3 minutes, I think (32-addresses for phase of a `BufRd`). But it may be possible to solve this with a `PlayBuf`, but you'd lose the nice cubic interpolation and there are a few calculations you'd have to do. We could automatically switch if the playback rate is `>= 1`.

The synth is this this, maybe you find a solution with `PlayBuf`?
```supercollider
(
var numChannels = ~dirt.numChannels;



	// write variants for different sample buffer sizes
	(1..SuperDirt.maxSampleNumChannels).do { |sampleNumChannels|

		var name = format("dirt_sample_%_%", sampleNumChannels, numChannels);

		SynthDef(name, { |out, bufnum, sustain = 1, begin = 0, end = 1, speed = 1, endSpeed = 1, freq = 440, pan = 0|

			var sound, rate, phase, sawrate, numFrames;

			// playback speed
			rate = Line.kr(speed, endSpeed, sustain) * (freq / 60.midicps);

			// sample phase
			// BufSampleRate adjusts the rate if the sound file doesn't have the same rate as the soundcard
			//phase =  Sweep.ar(1, rate * BufSampleRate.ir(bufnum)) + (BufFrames.ir(bufnum) * begin);

			numFrames = BufFrames.ir(bufnum);
			sawrate = rate * BufSampleRate.ir(bufnum) / (absdif(begin, end) * numFrames);
			phase = (speed.sign * LFSaw.ar(sawrate, 1)).range(begin,end) * numFrames;

			sound = BufRd.ar(
				numChannels: sampleNumChannels,
				bufnum: bufnum,
				phase: phase,
				loop: 0,
				interpolation: 4 // cubic interpolation
			);

			sound = DirtPan.ar(sound, numChannels, pan);

			Out.ar(out, sound)
		}, [\ir, \ir, \ir, \ir, \ir, \ir, \ir, \ir]).add;
	};
)
```

telephon:
[...]

PR #229, opened 2021-02-01
Stretch v2

bgold-cosmos:
Update of #149

### DIFF
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -100,6 +100,7 @@ DirtEvent {
 				delta * ~legato.value
 			} {
 				unitDuration = unitDuration ? delta;
+				if (~timescale.notNil) {unitDuration = unitDuration * ~timescale };
 				loop !? { unitDuration = unitDuration * loop.abs };
 			}
 		};
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -100,7 +100,6 @@ DirtEvent {
 				delta * ~legato.value
 			} {
 				unitDuration = unitDuration ? delta;
-				if (~timescale.notNil) {unitDuration = unitDuration * ~timescale };
 				loop !? { unitDuration = unitDuration * loop.abs };
 			}
 		};
@@ -111,6 +110,7 @@ DirtEvent {
 
 		~fadeTime = min(~fadeTime.value, sustain * 0.19098);
 		~fadeInTime = if(~begin != 0) { ~fadeTime } { 0.0 };
+		if (~timescale.notNil) {sustain = sustain * ~timescale };
 		~sustain = sustain - (~fadeTime + ~fadeInTime);
 		~speed = speed;
 		~endSpeed = endSpeed;
--- a/synths/core-modules.scd
+++ b/synths/core-modules.scd
@@ -53,7 +53,7 @@ this may be refacored later.
 
 				if(~timescale.notNil) {
 					instrument = ~stretchInstrument;
-					args = args.add(\timescale).add(~timescale).add(\timescalewin).add(~timescalewin ? 1);
+					args = args ++ [\timescale, ~timescale, \timescalewin, ~timescalewin ? 1];
 				} {
 					instrument = ~instrument;
 				};
--- a/synths/core-modules.scd
+++ b/synths/core-modules.scd
@@ -38,17 +38,32 @@ this may be refacored later.
 		if(~diversion.value.isNil) {
 			if(~buffer.notNil) {
 				// argumets could be omitted using getMsgFunc, but for making it easier to understand, we write them out
-				dirtEvent.sendSynth(~instrument,  [
-					bufnum: ~buffer,
-					sustain: ~sustain,
-					speed: ~speed,
-					freq: ~freq,
-					endSpeed: ~endSpeed,
-					begin: ~begin,
-					end: ~end,
-					pan: ~pan,
-					out: ~out
-				])
+				if(~timescale.notNil) {
+					dirtEvent.sendSynth(~stretchInstrument,  [
+						bufnum: ~buffer,
+						sustain: ~sustain,
+						speed: ~speed,
+						freq: ~freq,
+						endSpeed: ~endSpeed,
+						begin: ~begin,
+						end: ~end,
+						pan: ~pan,
+						timescale: ~timescale,
+						out: ~out
+					])
+				} {
+					dirtEvent.sendSynth(~instrument,  [
+						bufnum: ~buffer,
+						sustain: ~sustain,
+						speed: ~speed,
+						freq: ~freq,
+						endSpeed: ~endSpeed,
+						begin: ~begin,
+						end: ~end,
+						pan: ~pan,
+						out: ~out
+					])
+				}
 			} {
 				if(~instrument.isNil) {
 					"module 'sound': instrument not found: %".format(~s).postln
--- a/synths/core-modules.scd
+++ b/synths/core-modules.scd
@@ -37,33 +37,28 @@ this may be refacored later.
 	{ |dirtEvent|
 		if(~diversion.value.isNil) {
 			if(~buffer.notNil) {
-				// argumets could be omitted using getMsgFunc, but for making it easier to understand, we write them out
+				// arguments could be omitted using getMsgFunc, but for making it easier to understand, we write them out
+				args = [
+					bufnum: ~buffer,
+					sustain: ~sustain,
+					speed: ~speed,
+					freq: ~freq,
+					endSpeed: ~endSpeed,
+					begin: ~begin,
+					end: ~end,
+					pan: ~pan,
+					out: ~out
+				]
+
 				if(~timescale.notNil) {
-					dirtEvent.sendSynth(~stretchInstrument,  [
-						bufnum: ~buffer,
-						sustain: ~sustain,
-						speed: ~speed,
-						freq: ~freq,
-						endSpeed: ~endSpeed,
-						begin: ~begin,
-						end: ~end,
-						pan: ~pan,
-						timescale: ~timescale,
-						out: ~out
-					])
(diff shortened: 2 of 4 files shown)


bgold-cosmos:
One comment on this is that the timestretch algorithm used here doesn't really work with the workaround for issue #196.  So if you're timestretching long samples, it may sound weird, but I'm not sure when it becomes noticeable since the timestretching itself is going to sound weird.

yaxu:
I'm just trying to getting this running, I think it's just the `timescale` parameter that needs sending, right? 

Maybe we could call that parameter `stretch` instead? Tidal already has a function with that name, but it seems to only be a supporting function for `fit'`.

yaxu:
I've got it working now, sounding nice and gnarly! It would be nice to have a version that resizes it relative to the event duration (the `delta` param). Not sure what that could be called, maybe `pack`?

bgold-cosmos:
The parameter was originally called `stretch` and I can see in the commit history when it changed, but I don't remember at all the reason why.  I have a vague recollection of some discussion about possible name collisions in SuperCollider.

By your second comment, you mean a version that automatically stretches the sample so that it fills the event duration?

yaxu:
Ah OK, maybe @telephon remembers about the name collision. We could still call the parameter `stretch` on the tidal side, although it's nice to have things named the same everywhere.

And yes that's right. A bit like `legato`. In fact it could detect if `legato` is switched on, and if so change its behaviour so that the eventual duration in seconds is `legato * stretch * delta`.

telephon:
The parameter `stretch` is used in the event system of sclang for a different purpose, so we might want to avoid it.

telephon:
How should these parameters work for synths?

yaxu:
Perhaps we call it `timestretch` then, with the alias `stretch` in tidal.

For synths I guess we'd just multiply legato up using stretch and not use it otherwise? As synths are already stretchy by nature.

telephon:
> For synths I guess we'd just multiply legato up using stretch and not use it otherwise? As synths are already stretchy by nature.

yes. synths may still use the `timestretch` argument if they like, to add some inexplainably weird distortion :)

yaxu:
Oh I hadn't realised that synths could actually be timestretched with the same effect! That's strange
*edit* Oh re-reading I think I misunderstood. You're saying individual synths can add timestretch distortion? Heh

telephon:
> You're saying individual synths can add timestretch distortion? Heh

yes, they could do anything with that parameter, so that when you mix synths and samples in one pattern, they have some sort of analogy to each other.

classes/DirtEvent.sc
```diff
@@ -100,6 +100,7 @@ DirtEvent {
 				delta * ~legato.value
 			} {
 				unitDuration = unitDuration ? delta;
+				if (~timescale.notNil) {unitDuration = unitDuration * ~timescale };
```
telephon: I think this is not the right place to add the stretch. Better do it when the sustain has been fully calculated. We'll need to carefully test this.

```supercollider
		if (~timescale.notNil) { sustain = sustain * ~timescale };
		~sustain = sustain - (~fadeTime + ~fadeInTime);

```

synths/core-synths.scd
```diff
@@ -98,6 +98,60 @@ live coding them requires that you have your SuperDirt instance in an environmen
...
+			// playback speed
+			rate = Line.kr(speed, endSpeed, sustain) * (freq / 60.midicps);
+
+			// sample phase
+			// BufSampleRate adjusts the rate if the sound file doesn't have the same rate as the soundcard
+			//phase =  Sweep.ar(1, rate * BufSampleRate.ir(bufnum)) + (BufFrames.ir(bufnum) * begin);
+
+			numFrames = BufFrames.ir(bufnum);
+			// try windows that are 1/29th the width of the result, but not too small or large or it sounds weird
+			windowSize = numFrames * timescale / 29.0;
```
telephon: This means that the windows get larger the longer the sample is. That doesn't sound correct to me …

synths/core-synths.scd
```diff
@@ -98,6 +98,60 @@ live coding them requires that you have your SuperDirt instance in an environmen
...
+			windowSize = windowSize.clip(1000*timescale, BufSampleRate.ir(bufnum)/10);
```
telephon: if these values are guesses, it would be good if you add a comment how you guessed them

synths/core-synths.scd
```diff
@@ -98,6 +98,60 @@ live coding them requires that you have your SuperDirt instance in an environmen
...
+			timescaleStep = windowSize / timescale;
+			sawrate = rate * BufSampleRate.ir(bufnum) / (absdif(begin, end) * timescale * numFrames);
```
telephon, line 134: In general: the synth will become a bit more efficient I believe when you refactor so that divisions become multiplications as far as reasonable. This is in particular when the UGens involved are audio rate UGens. Just a suggestion!

telephon (changes requested):
just a few things, I think nothing complicated …

synths/core-modules.scd
```diff
@@ -53,7 +53,7 @@ this may be refacored later.
 
 				if(~timescale.notNil) {
 					instrument = ~stretchInstrument;
-					args = args.add(\timescale).add(~timescale);
+					args = args.add(\timescale).add(~timescale).add(\timescalewin).add(~timescalewin ? 1);
```
telephon: it might be more efficient (and better to read) to write:
```
args = args ++ [\timescale, ~timescale, \timescalewin, ~timescalewin ? 1]

```

synths/core-synths.scd
```diff
@@ -147,7 +149,7 @@ live coding them requires that you have your SuperDirt instance in an environmen
 			sound = DirtPan.ar(sound, numChannels, pan);
 
 			Out.ar(out, sound)
-		}, [\ir, \ir, \ir, \ir, \ir, \ir, \ir, \ir, \ir, \ir]).add;
+		}, [\ir, \ir, \ir, \ir, \ir, \ir, \ir, \ir, \kr, \ir, \ir]).add;
```
telephon, line 162: Making `pan` continuous is intentional? If yes, it is unrelated, so maybe it is a good idea to do this in a separate PR, checking all SynthDefs.

synths/core-synths.scd
```diff
@@ -116,9 +116,11 @@ live coding them requires that you have your SuperDirt instance in an environmen
...
 
 			numFrames = BufFrames.ir(bufnum);
-			// try windows that are 1/29th the width of the result, but not too small or large or it sounds weird
-			windowSize = numFrames * timescale / 29.0;
-			windowSize = windowSize.clip(1000*timescale, BufSampleRate.ir(bufnum)/10);
+			// Picking the right window size is a tricky thing, something around 2000 samples is usually
+			// OK for sounds of moderate length (around 0.5 to several seconds long).
+			// But this is also scaled below to be a bit smaller/larger based on desired playback rate.
+			// If it's still not good, you can use the "timescalewin" parameter to multiply the window size
+			windowSize = timescale.clip(0.1,2) * 0.05 * BufSampleRate.ir(bufnum) * timescalewin;
```
telephon, line 125: 1. you are clipping the value here, but in line 124 you are using it unclipped. Won't this lead to problems? Also in the line `if (~timescale.notNil) { sustain = sustain * ~timescale };` above, you are using the full sustain, while in the synthdef you are clipping it, while the sustain is still the same length.
2. Further down, there is a line:
```
windowIndex = phase - (phase.div(timescaleStep) - [1.0, 0.0] * timescaleStep);
```
Are you sure this is correct? Note that there is no operator precedence in supercollider. I'm unsure right now.

telephon (changes requested):
Thank you! Just some minor things …

synths/core-synths.scd
bgold-cosmos: This is just to match a change you made in #210 to dirt_sample_%_%, where pan is also control rate.

telephon: ok, good!

synths/core-synths.scd
bgold-cosmos: `windowSize` and `timescale` can be varied completely independently and the algorithm still works, but may sound odd for certain combinations.  I'm really clipping on windowSize values, but a "good" windowSize tends to scale somewhat with the length of the effect.  I'm not sure what you mean about the sustain.

The `windowIndex` calculation is correct, but unnecessarily obscure.  I added a variable to clarify it a bit (and it might save a repeated computation).

synths/core-modules.scd
bgold-cosmos: done!

synths/core-synths.scd
telephon: >I'm not sure what you mean about the sustain.

Let's say e.g. you are setting `# timescale 0.09` for a sample that is 1 second long. Then your playback synth will have a total duration (=sustain) of 0.09 seconds. But your window will be `(0.1 * 0.05)`. Or, when you set `# timescale 3`, your synth will be 3 seconds, but the window will be `(2 * 0.05)`. But also, in line 124 (`timescaleStep = windowSize / timescale;` e.g. for a very large timescale, the timescalestep will be very small, but the window won't get smaller because of the clipping.

I am not sure if all that is a problem, it just looks like a source of hard to trace errors. It might be easier to reason about when you limit the stretching before everything else?

telephon: Sorry to have left this dangling. I meant that because `timescaleStep` is clipped, the overlaps will become larger when `timescale < 0.1`. Could you check if that is really correct? If that is what you want, it might be good to make a comment.

bgold-cosmos: I'll take a look!  I have since noticed that a combo of frequency shift (`fshift`) and speed modification works pretty well for smaller amounts of stretching.

telephon (approved):

telephon:
Let me know if you want to add more polish.

Issue #199, opened 2020-05-11
Allow tidal to query superdirt for a list of samples / synths / effects?


yaxu:
We now have the `rms` capability, could we add a way for tidal to ask which samples / synths are available? Maybe effects too?

bgold-cosmos:
FWIW, on the SuperDirt end of things you can get samples, effects, and global effects with something like
```
// sample information
~dirt.postSampleInfo
// for a printout, or to get an array
~dirt.soundLibrary.bufferEvents.keys.asArray.sort

// registered "effects" should be in this list
~dirt.modules.collect({|x| x.name})

// global effects show up here
~dirt.orbits[0].globalEffects.collect({|x| x.name})
```

I expected
```
~dirt.soundLibrary.synthEvents.keys.asArray.sort
```
to list all the "synth" instruments but they don't seem to show up here, maybe I'm not registering them properly?  Usually I just make a SynthDef, and I think SuperDirt doesn't necessarily keep a lookup of every SynthDef since many of them may have nothing to do with SuperDIrt.

In the past to try to find all the "synths" I've defined I use this, but I think it may still grab things that are unrelated to SuperDirt
```
( // this bit should show all "instrument" synthDefs
var keys = SynthDescLib.global.synthDescs.keys;
keys = keys.select( {|k,i|
	var notSys = "system_".matchRegexp(k.asString).not;
	var notDirt = "dirt_".matchRegexp(k.asString).not;
	notSys && notDirt;
});
keys.asList.sort.do{|x| postln(x)};
nil;
)
```

telephon:
Yes, defining a synth event is optional, you just do it when you need specific parameters or transformations of parameters. The rest is just normal SynthDefs. As you show, you can easily get a list of all SynthDefs (`SynthDescLib.default.synthDescs`), but these are also others that won't work properly with tidal.

So some explicit bookkeeping is necessary for this.

yaxu:
This is related to https://github.com/tidalcycles/Dirt-Samples/issues/15 then. A general way of cataloguing samples and synths metadata, including #122 .

yaxu:
I'm talking with @lwlsn about this sort of thing.. Particularly how to make use of metadata about samples in Tidal patterns.

E.g. lets say I had a folder of short sounds, and wanted to play them in order of 'brightness':
```
d1 $ sound "vocalsnippets*16" # quality "spectral-centroid" saw
```
[...]

PR #234, opened 2021-02-27
Topic net responders

yaxu:


classes/SuperDirt.sc
```diff
@@ -284,14 +284,22 @@ SuperDirt {
...
-
 		netResponders.add(
+			OSCFunc({ |msg, time, tidalAddr|
+				var name = msg[1].asSymbol;
+				var synth = SynthDescLib.global.at(name);
+				var controls = synth.controls;
+				var controlNames = msg[1];
+				controls.do{arg control; controlNames = controlNames ++ " " ++ control.name};
+				tidalAddr.sendMsg("/dirt/synth-info/reply", controlNames);
+				controlNames.postln;
```
telephon: there is still a postln left in, probably we don't want this?

classes/SuperDirt.sc
```diff
@@ -284,14 +284,22 @@ SuperDirt {
...
 		};
 
-
 		netResponders.add(
+			OSCFunc({ |msg, time, tidalAddr|
+				var name = msg[1].asSymbol;
+				var synth = SynthDescLib.global.at(name);
+				var controls = synth.controls;
+				var controlNames = msg[1];
+				controls.do{arg control; controlNames = controlNames ++ " " ++ control.name};
```
telephon: `controls.do{ |control| ` is nicer, if you like.

classes/SuperDirt.sc
yaxu: oops, removed that now

classes/SuperDirt.sc
yaxu: done thanks!

classes/SuperDirt.sc
```diff
@@ -284,14 +284,21 @@ SuperDirt {
...
 
-
 		netResponders.add(
+			OSCFunc({ |msg, time, tidalAddr|
+				var name = msg[1].asSymbol;
+				var synth = SynthDescLib.global.at(name);
+				var controls = synth.controls;
+				var controlNames = msg[1];
+				controls.do{|control| controlNames = controlNames ++ " " ++ control.name};
+				tidalAddr.sendMsg("/dirt/synth-info/reply", controlNames);
```
telephon, line 294: One thing: don't you also want to send the default values over? Or is that not useful on the tidal side ...

yaxu: That might be useful for the human user to know. There could be more info, like the normal range of the parameter, whether it can be controlled via a bus and so on.. So maybe we should think a bit ahead and flesh out https://github.com/musikinformatik/SuperDirt/issues/199 a bit first.

(This PR came from a [question on the forum](https://club.tidalcycles.org/t/synthdef-argument-posting-function/2906?u=yaxu))

telephon: OK, I'm going over there …

yaxu:
Maybe it's OK to merge this as-is?

PR #236, opened 2021-03-23
midi: fix midi gain

### LINKED ISSUE #810: gain no longer affects MIDI velocity
`gain` patterns no longer seem to affect MIDI velocity in Tidal 1.7:

```
d1 $ gain "1 0.5 0.75" # s "midi" # note "c5
```

The above pattern results in all notes having the same velocity.

telephon:
This should fix https://github.com/tidalcycles/Tidal/issues/810


kindohm:
@telephon I did a quick test and this seems to be working now. Thank you!

PR #257, opened 2022-01-02
sound library: sort files in folder

### LINKED ISSUE #131: Inconsistent sorting of files

telephon:
This fixes  #131.

### DIFF
--- a/classes/DirtSoundLibrary.sc
+++ b/classes/DirtSoundLibrary.sc
@@ -15,6 +15,7 @@ DirtSoundLibrary {
 	var <>verbose = false;
 	var <>defaultEvent;
 	var <>doNotReadYet = false;
+	var <>sortFilesInFolders = true;
 
 	*new { |server, numChannels|
 		^super.newCopyArgs(server, numChannels).init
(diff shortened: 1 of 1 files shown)


telephon:
duplicated

PR #262, opened 2022-03-02
a hack to feed audio from MIDI or VST back into SuperDirt

kasparsj:
Hi, not sure how useful this is for others, but I tought I had share this SuperDirt MIDI hack.
It basically gives more control over MIDI instruments (hardware and VST) from TidalCycles, by feeding the audio back into the orbit, so that effects/panning can be applied.

### DIFF
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -140,7 +140,7 @@ DirtEvent {
 		args.asControlInput.flop.do { |each|
 			server.sendMsg(\s_new,
 				instrument,
-				-1, // no id
+				~synthID ? -1, // no id
 				1, // add action: addToTail
 				group, // send to group
 				*each.asOSCArgArray // append all other args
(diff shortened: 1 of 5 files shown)


telephon:
Thank you! I think there is no need to make the changes to the class interface. One DirtSoundLibrary can be shared between several instances of SuperDirt. 

You can reach the current orbit and the server from within the `play` function of a synth event, and using the `playInside` function field, you are passed the `DirtEvent` (see `hacks/function-hacks.scd` there are examples that might help)

Or is there something that can't be realised that way?

kasparsj:
> Thank you! I think there is no need to make the changes to the class interface. One DirtSoundLibrary can be shared between several instances of SuperDirt.
> 
> You can reach the current orbit and the server from within the `play` function of a synth event, and using the `playInside` function field, you are passed the `DirtEvent` (see `hacks/function-hacks.scd` there are examples that might help)
> 
> Or is there something that can't be realised that way?

Thanks for the comments. Yes, that makes sense. I have updated the hack accordingly. Just 1 little change to DirtEvent class now required.

telephon:
Ah yes, very much better. Let me know when you need another round of comments.

telephon:
@kasparsj sorry for not responding – I forgot about it. This is interesting, I would like to find a less intrusive way to do it. I suppose that the best way to run VST Plugins would be in global effects, because they do not need to be started at each onset. But I don't understand everything you are doing here …

Issue #273, opened 2022-06-26
playing samples in tune using pitch metadata


ahihi:
currently, playing samples tonally is quite cumbersome unless they are recorded at some standard pitch (C seems to be a common choice). what if we could tag sample files with a base pitch, have SuperDirt read that information and make it automatically repitch the samples to match incoming notes?

in fact i have already [implemented this](https://github.com/ahihi/SuperDirt/tree/wav-pitch-meta) :) here is how it works:

1. pitch metadata is added to WAV files in the form of MIDI unity note / MIDI pitch fraction fields of a [`smpl` chunk](https://www.recordingblogs.com/wiki/sample-chunk-of-a-wave-file), using a utility such as my [pitcheon](https://github.com/ahihi/pitcheon)
2. SuperDirt obtains the pitch metadata by parsing `Soundfile.readHeaderAsString`[^1]
3. for events containing a `tune` argument with value 1, frequency and sample duration are scaled relative to the obtained pitch metadata (`tune` defaults to 0, preserving the old behavior)

or, from the musician's perspective:

1. i have a bunch of one-shot samples recorded at wildly mismatched pitches; i do `d1 $ n "<0 1 2 3 4>*8" # s "mysamples"` and it sounds terrible!
2. i run pitcheon on my samples
3. i restart SuperDirt, do `d1 $ n "<0 1 2 3 4>*8" # s "mysamples" # pF "tune" 1`, everything plays in a beautiful consonance of C, audience erupts into applause, i am booked for numerous high-paying gigs

...anyway, would you be interested in upstreaming this feature? my implementation probably has room for improvement but i would be happy to work out the details in a PR.

[^1]: this string is basically a log written by libsndfile. it would be nicer to access the raw data, not only for performance and regex-matching-ugly reasons but because libsndfile does some questionable math on the pitch info which i have to [undo](https://github.com/ahihi/SuperDirt/blob/da768943e5c13f1a1b86cb630d3fc41c31320bfb/classes/DirtSoundLibrary.sc#L255).

telephon:
Good thing! Perhaps it could be quite a bit simpler? One could just add a multiplier that is derived from `baseFreq / 440` and multiplied at the end?

ahihi:
closing this now that the PR is merged 🥳

PR #274, opened 2022-07-05
sample repitching using smpl chunk metadata and tune argument

ahihi:
implements #273, now with the simplified approach suggested by @telephon.

### DIFF
--- a/classes/DirtOrbit.sc
+++ b/classes/DirtOrbit.sc
@@ -176,7 +176,8 @@ DirtOrbit {
 			~n = \none; // sample number or note
 			~octave = 5;
 			~midinote = #{ ~note ? ~n + (~octave * 12) };
-			~freq = #{ ~midinote.value.midicps };
+			~baseFreq = #{ 60.midicps };
+			~freq = #{ ~midinote.value.midicps * 60.midicps / ~baseFreq.value };
 			~dur = 1.0;
 			~delta = #{ ~dur.value };
 
--- a/classes/DirtOrbit.sc
+++ b/classes/DirtOrbit.sc
@@ -176,8 +176,8 @@ DirtOrbit {
 			~n = \none; // sample number or note
 			~octave = 5;
 			~midinote = #{ ~note ? ~n + (~octave * 12) };
-			~baseFreq = #{ 60.midicps };
-			~freq = #{ ~midinote.value.midicps * 60.midicps / ~baseFreq.value };
+			~baseFreqToFreqRatio = #{ 1.0 };
+			~freq = #{ ~midinote.value.midicps * ~baseFreqToFreqRatio.value };
 			~dur = 1.0;
 			~delta = #{ ~dur.value };
 
--- a/classes/DirtOrbit.sc
+++ b/classes/DirtOrbit.sc
@@ -176,8 +176,7 @@ DirtOrbit {
 			~n = \none; // sample number or note
 			~octave = 5;
 			~midinote = #{ ~note ? ~n + (~octave * 12) };
-			~baseFreqToFreqRatio = #{ 1.0 };
-			~freq = #{ ~midinote.value.midicps * ~baseFreqToFreqRatio.value };
+			~freq = #{ ~midinote.value.midicps };
 			~dur = 1.0;
 			~delta = #{ ~dur.value };
 
--- a/classes/DirtSoundLibrary.sc
+++ b/classes/DirtSoundLibrary.sc
@@ -235,7 +235,7 @@ DirtSoundLibrary {
 	}
 
 	readMetaData { |path|
-		^this.readSmplMetaData(path);
+		^this.readSmplMetaData(path)
 	}
 
 	// read the `smpl` chunk of a wave file, containing metadata such as pitch
@@ -272,7 +272,7 @@ DirtSoundLibrary {
 				midinote: midinote,
 				baseFreqToMetaFreqRatio: (60 - midinote).midiratio
 			)
-		};
+		}
 	}
 
 	/* access */
--- a/classes/DirtSoundLibrary.sc
+++ b/classes/DirtSoundLibrary.sc
@@ -47,14 +47,7 @@ DirtSoundLibrary {
 		event = this.makeEventForBuffer(buffer, metaData);
 		buffers[name] = buffers[name].add(buffer);
 		bufferEvents[name] = bufferEvents[name].add(event);
-		metaData !? {
-			if(metaDataEvents[name].isNil) {
-				metaDataEvents[name] = Dictionary[];
-			};
-			index = buffers[name].size - 1;
-			metaDataEvents[name].put(index, metaData);
-		};
-		if(verbose) { "new sample buffer named '%':\n%\n\n".postf(name, event) };
+		metaData !? { this.prPutMetaData(name, buffers[name].size - 1, metaData) };
 	}
 
 	addSynth { |name, event, appendToExisting = false, useSynthDefSustain = false, metaData|
@@ -70,7 +63,7 @@ DirtSoundLibrary {
 		if(event[\hash].isNil) { event[\hash] = name.identityHash };
 		if(useSynthDefSustain) { this.useSynthDefSustain(event) };
 		synthEvents[name] = synthEvents[name].add(event);
-		metaData !? { metaDataEvents[name] = metaDataEvents[name].add(metaData) };
+		metaData !? { this.prPutMetaData(name, synthEvents[name].size - 1, metaData) };
 		if(verbose) { "new synth named '%':\n%\n\n".postf(name, event) };
 	}
 
@@ -427,6 +420,10 @@ DirtSoundLibrary {
 		synthEvents = synthEvents.copy;
 	}
 
-
-
+	prPutMetaData { |name, index, metaData|
+		if(metaDataEvents[name].isNil) {
+			metaDataEvents[name] = Dictionary[];
+		};
+		metaDataEvents[name].put(index, metaData);
+	}
 }
--- a/classes/DirtSoundLibrary.sc
+++ b/classes/DirtSoundLibrary.sc
@@ -44,7 +44,7 @@ DirtSoundLibrary {
 			"\nreplacing '%' (%)\n".postf(name, buffers[name].size);
 			this.freeSoundFiles(name);
 		};
-		event = this.makeEventForBuffer(buffer);
+		event = this.makeEventForBuffer(buffer, metaData);
 		buffers[name] = buffers[name].add(buffer);
 		bufferEvents[name] = bufferEvents[name].add(event);
 		metaData !? {
@@ -250,15 +250,18 @@ DirtSoundLibrary {
 				noteStr = headers.findRegexp(noteRe)[1][1];
 				fractStr = headers.findRegexp(fractRe)[1][1];
 			});
-			["noteStr", noteStr].postln;
-			["fractStr", fractStr].postln;
 			note = noteStr.asInteger;
-			fract = fractStr.asInteger/128.0;
-			["note", note].postln;
-			["fract", fract].postln;
+			fract = fractStr.asFloat;
+			// as of 2022-06-24, the libsndfile calculation that gives the pitch fraction seems to be wrong...
+			// https://github.com/libsndfile/libsndfile/blob/33e765ccba9a0eb225694fdbf9e299683a8338ee/src/wav.c#L1399
+			// basically, all values other than 0 are inverted and scaled incorrectly.
+			// here we assume this is the case and correct the math.
+			// TODO: maybe have a flag to skip this, to support a potential fixed libsndfile?
+			if(fract > 0) {
+				fract = fract.reciprocal * 0.5;
+			};
 			midinote = note + fract;
-		} { |e| e.postln; nil; };
-		["midinote", midinote].postln;
+		} { |e| e.reportError; nil; };
 		^if(midinote.notNil) { (midinote: midinote) };
 	}
 
@@ -313,7 +316,7 @@ DirtSoundLibrary {
 		}
 	}
 
-	makeEventForBuffer { |buffer|
+	makeEventForBuffer { |buffer, meta|
 		var baseFreq = 60.midicps;
 		^(
 			buffer: buffer.bufnum,
@@ -323,7 +326,14 @@ DirtSoundLibrary {
 			stretchInstrument: this.stretchInstrumentForBuffer(buffer),
 			bufNumFrames: buffer.numFrames,
 			bufNumChannels: buffer.numChannels,
-			unitDuration: { buffer.duration * baseFreq / ~freq.value },
+			baseFreq: {
+				if(~tune > 0) {
+					meta !? _.midinote !? _.midicps
+				} ? baseFreq
+			},
+			unitDuration: {
+				buffer.duration * baseFreq / ~freq.value
+			},
 			hash: buffer.identityHash,
 			note: 0
 		)
--- a/classes/DirtSoundLibrary.sc
+++ b/classes/DirtSoundLibrary.sc
@@ -33,7 +33,7 @@ DirtSoundLibrary {
 	}
 
 	addBuffer { |name, buffer, appendToExisting = false, metaData|
-		var event;
+		var event, index;
 		if(buffer.isNil) { Error("tried to add Nil to buffer library").throw };
 		if(synthEvents[name].notNil) {
 			"a synth event with that name already exists: %\nSkipping...".format(name).warn;
@@ -48,7 +48,6 @@ DirtSoundLibrary {
 		buffers[name] = buffers[name].add(buffer);
 		bufferEvents[name] = bufferEvents[name].add(event);
 		metaData !? {
-			var index;
 			if(metaDataEvents[name].isNil) {
 				metaDataEvents[name] = Dictionary[];
 			};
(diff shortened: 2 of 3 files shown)


classes/DirtSoundLibrary.sc
```diff
@@ -222,6 +233,43 @@ DirtSoundLibrary {
 		^Buffer.readWithInfo(server, path, onlyHeader: doNotReadYet)
 	}
 
+	readMetaData { |path|
+		^this.readSmplMetaData(path);
+	}
+
+	readSmplMetaData { |path|
```
telephon, line 237: readSmplMetaData -> readSoundFileMetaData

classes/DirtSoundLibrary.sc
```diff
@@ -198,7 +205,8 @@ DirtSoundLibrary {
 			try {
 				var buf = this.readSoundFile(filepath);
 				if(buf.notNil) {
-					this.addBuffer(name, buf, appendToExisting);
+					var metaData = this.readMetaData(filepath);
```
telephon: please move variable definition outside of if statement for efficiency. Then in can be inlined by the compiler.

classes/DirtSoundLibrary.sc
```diff
@@ -44,10 +44,17 @@ DirtSoundLibrary {
 			"\nreplacing '%' (%)\n".postf(name, buffers[name].size);
 			this.freeSoundFiles(name);
 		};
-		event = this.makeEventForBuffer(buffer);
+		event = this.makeEventForBuffer(buffer, metaData);
 		buffers[name] = buffers[name].add(buffer);
 		bufferEvents[name] = bufferEvents[name].add(event);
-		metaData !? { metaDataEvents[name] = metaDataEvents[name].add(metaData) };
+		metaData !? {
+			var index;
```
telephon: move out of fuction, so if can be inlined. Tip: 
```
// do this in your config
LanguageConfig.postInlineWarnings = true;
```
then you'll be informed …

classes/DirtOrbit.sc
```diff
@@ -176,7 +176,8 @@ DirtOrbit {
 			~n = \none; // sample number or note
 			~octave = 5;
 			~midinote = #{ ~note ? ~n + (~octave * 12) };
-			~freq = #{ ~midinote.value.midicps };
+			~baseFreqToFreqRatio = #{ 1.0 };
```
telephon: no need to wrap this in a function. This can simply be a number, can't it?

classes/DirtOrbit.sc
```diff
@@ -176,7 +176,8 @@ DirtOrbit {
...
+			~freq = #{ ~midinote.value.midicps * ~baseFreqToFreqRatio.value };
```
telephon: no need to call `.value`?

classes/DirtSoundLibrary.sc
```diff
@@ -222,6 +233,43 @@ DirtSoundLibrary {
...
+				// https://github.com/libsndfile/libsndfile/blob/33e765ccba9a0eb225694fdbf9e299683a8338ee/src/wav.c#L1399
+				// basically, all values other than 0 are inverted and scaled incorrectly.
+				// here we assume this is the case and correct the math.
+				// TODO: maybe have a flag to skip this, to support a potential fixed libsndfile?
+				if(fract > 0) {
+					fract = fract.reciprocal * 0.5;
+				};
+				midinote = note + fract;
+			};
+		} { |e| e.reportError; nil; };
```
telephon: no need to return `nil` and no need for semicolon in the return line.

so: `{ |e| e.reportError };`

classes/DirtSoundLibrary.sc
```diff
@@ -274,7 +322,7 @@ DirtSoundLibrary {
 		}
 	}
 
-	makeEventForBuffer { |buffer|
+	makeEventForBuffer { |buffer, meta|
```
telephon: meta -> metaData

classes/DirtSoundLibrary.sc
```diff
@@ -284,6 +332,11 @@ DirtSoundLibrary {
 			stretchInstrument: this.stretchInstrumentForBuffer(buffer),
 			bufNumFrames: buffer.numFrames,
 			bufNumChannels: buffer.numChannels,
+			baseFreqToFreqRatio: {
+				if(~tune.notNil && { ~tune > 0 }) {
```
telephon: for efficiency use `and:` instead of `&`.

telephon: Something like this perhaps?
```supercollider
var baseFreqToFreqRatio;
baseFreqToFreqRatio = if(~tune.notNil and: { ~tune > 0 } and: { metaData.notNil }) {
						 _.baseFreqToFreqRatio
				} ? 1.0;

<...>

metaData: metadata,
baseFreqToFreqRatio: baseFreqToFreqRatio
```

`tune` is an ambiguous variable, perhaps `tuneToMetaData` – this is long, but it won't have to be written a lot has it? Where is it set and why is it `> 0`?

telephon (changes requested):
Great, looks already very good. There are some suggestions, let me know what you think ...

ahihi:
thank you for the thoughtful review, i am learning a lot! :D

classes/DirtSoundLibrary.sc
ahihi: the enclosing function is called from `addBuffer` but `~tune` doesnt exist yet at this point, it is a patternable argument (that is why i have the wrapping function). but thinking about it some more, the whole ratio is only relevant to the `sound` module, so actually i think it makes more sense to do the math in the module code instead..

youre right about the ambiguity, but `tuneToMetaData` does feel quite long compared to other argument names. how about `metatune`? (imagining a naming convention `metafoo` for "do foo using metadata")

`> 0` maybe should be `!= 0`. right now i am thinking of this as a boolean but also might experiment with other tuning modes later, making valid values `0,1,2,…` or possibly floats (`0.5` to tune 50% of the way to the target frequency? 😄 idk)

classes/DirtSoundLibrary.sc
ahihi: my thought was that `readSmplMetaData` specifically reads metadata stored in the `smpl` chunk, while the caller `readMetaData` might in the future also read other kinds and combine it all. `readSoundFileMetaData` sounds to me general enough that the two functions could just be merged into one

ahihi:
let me know what you think about this version!

classes/DirtSoundLibrary.sc
telephon: ah yes, that is good. I didn't know that the chunk was called `smpl`. Maybe worth a comment?

synths/core-modules.scd
```diff
@@ -35,15 +35,19 @@ this may be refacored later.
 
 ~dirt.addModule('sound',
 	{ |dirtEvent|
-		var instrument, args;
+		var instrument, freq, args;
 		if(~diversion.value.isNil) {
 			if(~buffer.notNil) {
+				freq = ~freq;
+				if(~tune.notNil and: { ~tune != 0 }) {
+					freq = freq * (~metaData !? _.baseFreqToFreqRatio ? 1.0);
```
telephon: since `metaData` is a Dictionary, better write:
```
freq = freq * (~metaData !? _[\baseFreqToFreqRatio] ? 1.0);
```
Or, in a way, it is clearer and more efficient:
```
if(~tune.notNil and: { ~tune != 0 } and: { ~metaData.notNil }) {
freq = freq * (metaData[\baseFreqToFreqRatio] ? 1.0) }
}
```

telephon:
> thank you for the thoughtful review, i am learning a lot! :D

great :) The code is already very good, so it is just some tweaking.

ahihi:
sorry for the long delay! ive now addressed your previous comments.

still TBD is the name of the argument for enabling this feature :) do you think `metatune` is acceptable? (ref. https://github.com/musikinformatik/SuperDirt/pull/274#discussion_r917135727)

ahihi:
while using this in my own livecoding, i noticed incorrect durations for some tuned notes.. i fixed this before but i think https://github.com/musikinformatik/SuperDirt/pull/274/commits/b53a29f89174cda02429a82e77a7c6a92e84473c broke it again. will take a look at it soon

ahihi:
note durations are now fixed. in the absence of further comments on the argument name i also took the liberty of renaming it to `metatune`. unless you would like some other changes i believe this should be ready to merge now!

telephon (changes requested):
I think it would be better to keep the dynamic value construction in the modules, but put the static stuff in DirtSoundLibrary.

So `baseFreqToMetaFreqRatio` in `DirtSoundLibrary` and `if(~metatune.notNil) {` etc. in `core-modules.scd`.

Then people can tweak the usage live and the sound library doesn't have to know about how the parameter is used.

ahihi:
i agree it would be better, but i had some trouble making it work that way (the commit was sitting on my computer for a while so i dont remember the details anymore). i will give it another shot

ahihi:
right, the problem is that by the time the module function is invoked, DirtEvent.calcTimeSpan has already happened and it is too late to modify unitDuration.

i wrote down this pseudocode summary of the event construction control flow, in case it is helpful:

```supercollider
DirtOrbit.init {
    makeDefaultParentEvent {
        defaultParentEvent = Event.make {
            ...
            ~midinote = #{ ~note ? ~n + (~octave * 12) }
            ~freq = #{ ~midinote.value.midicps }
            ...
        }
    }
}

DirtSoundLibrary.loadSoundFile {
    readSoundFile
    readMetaData
    addBuffer {
        bufferEvents[name] = makeEventForBuffer {
            builds event object, includes:
            - baseFreqToMetaFreqRatio (constant)
            - unitDuration (lambda, calls ~freq.value)
        }
    }
}

SuperDirt.connect {
    defines playFunc { |msg, ...|
        event = (latency: ...)
        event.putPairs(msg[1..])
        receiveAction.(event) // receiveAction is nil by default -> no-op
        DirtEvent(..., event).play {
            event.parent = orbit.defaultParentEvent
            event.use {
                // ignoring ~diversion != nil case
                mergeSoundEvent {
                    currentEnvironment.proto = orbit.dirt.soundLibrary.getEvent
                }
                orderTimeSpan // swaps ~begin and ~end if ~speed < 0
                calcTimeSpan {
                    applies ~unitDuration
                }
                playSynths {
                    modules.do(_.(this)) {
                        sound module called
                    }
                }
            }
        }
    }
    registers playFunc as OSC handler for /play2
}
```

telephon:
Ah yes, I see.

Maybe there is a better way to do this.

Can you try this:

```
~dirt.set(\freq, { ~midinote.value.midicps * ~metaTuneRatio.value }, \metaTuneRatio, 1.0);
```

and then you can just set the  `metaTuneRatio` in the SoundLibrary, I think.

If that works, we can extend the frequency calculation model of superdirt in `DirtOrbit` accordingly.


**COMMENT: actually, thinking about it, it is better as it is in this pull request. Saves some energy and keeps things together.**
I add some minor comments.

classes/DirtSoundLibrary.sc
```diff
@@ -222,6 +234,46 @@ DirtSoundLibrary {
 		^Buffer.readWithInfo(server, path, onlyHeader: doNotReadYet)
 	}
 
+	readMetaData { |path|
+		^this.readSmplMetaData(path);
```
telephon: you can leave out the semicolon for the return statement, this is nicer.

classes/DirtSoundLibrary.sc
```diff
@@ -44,10 +44,16 @@ DirtSoundLibrary {
...
+		event = this.makeEventForBuffer(buffer, metaData);
 		buffers[name] = buffers[name].add(buffer);
 		bufferEvents[name] = bufferEvents[name].add(event);
-		metaData !? { metaDataEvents[name] = metaDataEvents[name].add(metaData) };
+		metaData !? {
+			if(metaDataEvents[name].isNil) {
+				metaDataEvents[name] = Dictionary[];
+			};
+			index = buffers[name].size - 1;
+			metaDataEvents[name].put(index, metaData);
```
telephon: What is wrong with 
```
metaData !? { metaDataEvents[name] = metaDataEvents[name].add(metaData) };
```
? It should be right …

classes/DirtSoundLibrary.sc
```diff
@@ -284,7 +337,15 @@ DirtSoundLibrary {
 			stretchInstrument: this.stretchInstrumentForBuffer(buffer),
 			bufNumFrames: buffer.numFrames,
 			bufNumChannels: buffer.numChannels,
-			unitDuration: { buffer.duration * baseFreq / ~freq.value },
+			baseFreqToMetaFreqRatio: baseFreqToMetaFreqRatio,
+			metaTuneRatio: {
```
telephon: - [ ] maybe rename `metaTuneRatio` to `metaDataTuneRatio`? `metaTune` is ok, because it is short and we use it in tidal.

telephon (changes requested):

classes/DirtSoundLibrary.sc
telephon: your code replaces the the last item, instead of appending. Also, `metaDataEvents[name]`  should be an array, I think.

ahihi: my intention with the dictionary was to keep the metadata collection sparse rather than filling an array with nils in the case that most samples lack smpl metadata. in fact this does not replace any item - the dictionary key `index` is calculated from `buffers[name]` (array) but the collection that is written to is `metaDataEvents[name]` (dictionary), which at this point has no item at `index`.

telephon: Ah I see, it is a dictionary. Then it might be better to use the buffer itself as the key: `metaDataEvents[name].put(buffers[name], metaData);` (if there is not more than one metadata entry per buffer name).

ahihi: > (if there is not more than one metadata entry per buffer name)

there can be, since a buffer name represents a directory of samples, each of which can have different metadata (or no metadata).

ahihi:
here is an example that perhaps helps clarify how this works in practice.

samples used: https://share.pulusound.fi/metatune-samples.zip

contains a subdirectory `foo` with four samples. this is their smpl metadata:

```shell
$ find samples -iname '*.wav' -exec pitcheon '{}' ';'

samples/foo/00 hypersaw.wav
  1 smpl chunk(s)
    [0] note: 50.000 / freq: 146.832Hz / smpl: (50, 0)

samples/foo/01 bell.wav
  1 smpl chunk(s)
    [0] note: 71.410 / freq: 505.719Hz / smpl: (71, 1760936591)

samples/foo/02 piano.wav
  0 smpl chunk(s)

samples/foo/03 pluck.wav
  1 smpl chunk(s)
    [0] note: 68.000 / freq: 415.305Hz / smpl: (68, 0)
```

note that `02 piano.wav` has none, and all the others are have different pitches.

superdirt startup code:

```supercollider
(
s = Server.local;
s.options.sampleRate_(44100);
s.options.memSize = 1024 * 1024;
s.options.numBuffers = 1024 * 256;
s.options.numWireBufs = 128;
s.options.maxNodes = 1024 * 32;
s.latency_(0.205);
s.newBusAllocators;

s.waitForBoot {
	~dirt = SuperDirt(2, s);
	~dirt.loadSoundFiles(thisProcess.nowExecutingPath.dirname +/+ "samples/*");
	s.sync;
	~dirt.start(outBusses: 0 ! 12);
};
)
```

after this, `~dirt.soundLibrary.metaDataEvents` looks like:

```supercollider
IdentityDictionary[
  (foo -> Dictionary[
    (0 -> ( 'baseFreqToMetaFreqRatio': 1.7817974362766, 'midinote': 50.0 )),
    (1 -> ( 'baseFreqToMetaFreqRatio': 0.51733355006402, 'midinote': 71.4100000656 )),
    (3 -> ( 'baseFreqToMetaFreqRatio': 0.6299605249486, 'midinote': 68.0 ))
  ])
]
```

tidal pattern:

```haskell
d1
  $ n "[0 1 2 3]*2"
  # s "foo"
  # legato 1
  # pF "metatune" 0
```

compare values `0` and `1` for `metatune`. the piano sample is unaffected due to the aforementioned lack of metadata.

classes/DirtSoundLibrary.sc
telephon: OK, still I don't think you need to calculate an index. You could just take the buffer object itself as key, maybe.

ahihi: personally i tend to prefer keys to be simple values that require minimal dependencies to construct, as i find it makes debugging easier. it also gives a nice symmetry between `buffers[name][i]`, `bufferEvents[name][i]` and `metaDataEvents[name][i]`. but i am open to changing it if you think the buffer object would be better. how is equality between Buffers defined, object identity?

telephon: sorry for my late reply. Yes, I agree that the symmetry is good. Of course there is one hidden asymmetry that the underlying object is an array for bufferEvents and a dictionary for metaDataEvents. But we'd have that anyway, so I think it is fine. I'll look thtough the code.

telephon (commented):
Why is there metaData for each buffer in the buffer array, but only one metaData item for the whole array of synthEvents? This makes the structure of the `metaDataEvents` a little weird, because for buffers it is dictionaries with indices to dictionaries, and for synths it is just dictionaries, but not with indices. Maybe I haven't understood something, but it looks that way …

Apart from that it is good, I think!

ahihi:
good question! i forgot about the synth case. it might be best to factor out the dictionary-manipulating code into a function that is used by both `addBuffer` and `addSynth`.

i also recently noticed metadata not getting correctly updated when reloading samples from a folder which has been loaded before, will look into that too.

ahihi:
how about this?

> i also recently noticed metadata not getting correctly updated when reloading samples from a folder which has been loaded before, will look into that too.

i could not reproduce this anymore, maybe i just did something wrong last time.

telephon:
Thank you! This is good.

Issue #254, opened 2021-11-28
`accelerate` not working with synths


cleary:
```
d1 $ note "c" # s "superchip" # accelerate "-1"
```
has previously accelerated the note in the superchip synth (I've also tried a bunch of other synths, samples **do** work).

My testing environments use:
 - Ubuntu 20.04
 - SuperDirt branches tested: `develop` and `master`
 -  Supercollider `3.11.2+repack-focal1` (manual backport) and  `3.10.0+repack-1ubuntu2` (from repos)

Other users who've verified the problem use both Windows and OSX, on the superdirt `master` branch.

Some users have reported it working (on SC 3.12.1, SD `master`, arch and win10)

The last record of it working I have was a livestream I did in Sept 2020:
https://youtu.be/_uASsk8nHvQ?t=446

A stream from the following month and it seems to not be working (I think I assumed I was doing something wrong, but forgot to follow up):
https://youtu.be/kliiBKWWy4w?t=233

yaxu:
Perhaps it was this change:
https://github.com/musikinformatik/SuperDirt/pull/244

bgold-cosmos:
superchip uses a UGen that emulates an old home computer synth chip, and that UGen takes a frequency divider as input for pitch control just like the original hardware.  Anyway, I'm not surprised if some change to freq scaling broke the conversion.  

But accelerate should certainly be doing something with other synths, I'll take a look and see if I can figure out what's going on.

ndr-brt:
Just to report, it works on my machine:
* Ubuntu 21.10
* Superdirt latest commit on `develop` branch
* Supercollider 3.12.1

When we introduced `DirtFreqScale` I had another computer and it worked on that too, I can't recall which supercollider version it was installed

PR #247, opened 2021-06-03
Apply DirtFreqScale to default synths

ndr-brt:
As #242

PR #280, opened 2022-12-01
Set DirtFreqScale default factor to 1

wimmers:
It seems that all applications of `DirtFreqScale` in `library/default-synths-extra.scd` are without effect: the parameter `speedFreq` is not passed, and it's default value is 0, effectively deactivating frequency scaling.

As `DirtFreqScale` is used nowhere else, I propose to change the default value of `speedFreq` to 1.

Per #247, this should recover the old behavior of `accelerate`, fixing #254.

### DIFF
--- a/classes/SuperDirtUGens.sc
+++ b/classes/SuperDirtUGens.sc
@@ -161,7 +161,7 @@ DirtFreqScale : UGen {
 	*kr { |speed = 1, accelerate = 0, sustain = 1, speedFreq|
 		var speedTerm;
 		speed = speed.abs;
-		speedFreq = speedFreq ?? { \speedFreq.ir(0) };
+		speedFreq = speedFreq ?? { \speedFreq.ir(1) };
 		speedTerm = Line.kr(speed, speed * (accelerate + 1), sustain);
 		// linear interpolation between a factor of 1 (speedFreq = 0) and of speedTerm (speedFreq = 1)
 		^speedFreq * (speedTerm - 1) + 1


telephon:
I had expected that one would want to switch it on selectively, by writing  something like:

```haskell
d1 $ note "c" # s "superchip" # accelerate "-1" # speedFreq 1`
```

or by setting a default value:

```supercollider
~dirt.set(\speedFreq, 1)
```

but I see the point that when you write accelerate, you imply it is on.

wimmers:
Thank you, I understand that intuition.
That means it could also be viewed as a documentation problem on the Tidal side.

On the other hand, there may still be quite some users that see this paramter in someone else's code, and when they just want to start playing with it, they will be confused by this behavior if they don't check the docs first.
So maybe it would be more beginner-friendly this way?

telephon:
Makes sense. Thinking about it, what I am worried about a little bit is that changing this will change many people's music: whoever has used `accelerate` together with synths (perhaps mixing samples and synths) may be surprised by it.

cleary:
Just following up on this (it's been a while) - 

> Makes sense. Thinking about it, what I am worried about a little bit is that changing this will change many people's music: whoever has used `accelerate` together with synths (perhaps mixing samples and synths) may be surprised by it.

Personal perspective, I've used `accelerate` extensively with synths myself, and am currently unable to use my old patterns because `accelerate` isn't working at all ... which surprised me too :(
I think having `accelerate` do *something* is more important than the consideration of people who are relying on a glitch essentially, ie for it to do nothing (even though it's defined, and listed as an effect in all the Tidal documentation)

telephon:
Could you paste a few examples with a comment how they should work? I need to understand better what the idea was, and you are the best person to know! Thanks ...

cleary:
@telephon 
Here's an example using `supergong`:
[Line 207](https://github.com/cleary/livecode/blob/478b9a54e3631aef0fc7f08aa13424ad4fd00478/sets/20200809_newmoon_marathon.tidal#L207C9-L208C1) from this New Moon set in Aug 2020:  
```
let melody = note (scale "ionian" "[< 1 0>  <8 7>] <[ ~ 2 3 4] [~ 2 1 -1]> < ~ -2 ~ 0 > ~") |+ note "<0!4 5!4>"
in
d1 $ almostNever (# accelerate 2) 
  $ jux (off 0.25 (# gain 0.3)) $ almostNever (jux (# squiz (sine + 1 ))) 
  $ stut 4 0.5 0.2 $ slow 4 $ melody 
  # sound "supergong" 
  # sustain 2 # octave 6 # gain 1.1
```
Here's the video from the marathon, at the time that this line is used - you can hear the higher pitched gong notes sliding (accelerating) up occasionally:
https://youtu.be/_tgKw2WdJTo?si=jgY6LWBB3xLcLK10&t=840

The current state of SuperDirt is that accelerate is completely ineffective (but it's listed everywhere in the [synth documentation](http://tidalcycles.org/docs/reference/synthesizers/)) - this implies to me it's a regression and a bug

telephon:
Thank you – a nice example to listen to!

Now I am checking the behaviour in a few lines.

```

{ DirtFreqScale.kr(speed: 1, accelerate:  1, sustain:  0.1, speedFreq: 0).poll; 0 }.play; // -> 1
{ DirtFreqScale.kr(speed: 1, accelerate:  1, sustain:  0.1, speedFreq: 1).poll; 0 }.play; // -> 2

```

So there is something wrong with the original implementation which your changed default doesn't fix: all synths would start  out an octave higher when you use accelerate.

I've moved this for discussion: https://club.tidalcycles.org/t/accelerate-for-synths/4889

telephon (approved):

PR #285, opened 2023-01-30
Send RPN Null after NRPN

### LINKED ISSUE #283: Should we send RPN Null after NRPN?
Also at [MIDI NRPN Null Needed? How? @ Tidal club](https://club.tidalcycles.org/t/midi-rpn-null-needed-how/4488).

My synth manual and many other manuals contain a statement like this "Once an NRPN parameters
has been specified, all Data Entry messages received on that channel will modify the value of that parameter. To prevent accidents, it is recomended that you set RPN Null (RPN Number = 7FH/7FH) when you have finished setting the value of the desired parameter."

Jwaldmann didn't face issues when not following this approach, so it might not be needed. But I feel it might anyway be best to follow the recommendation.

Should we do as recommended and follow the approach explained here http://www.philrees.co.uk/nrpnq.htm? It would be simple to add here:
https://github.com/musikinformatik/SuperDirt/blob/f1c4ff923eeabcc8350c45433ef8623d0a624263/classes/DirtEventTypes.sc#L89-L101

Zalastax:
Closes #283

I have used MIDI-OX to monitor the events being sent and it looks good. I have tested this toward my JD-Xi and it receives and acts on the NRPN changes as expected.

### DIFF
--- a/classes/DirtEventTypes.sc
+++ b/classes/DirtEventTypes.sc
@@ -97,6 +97,8 @@ DirtEventTypes {
 						midiout.control(chan, 98, nrpnLSB);
 						midiout.control(chan, 6,  valMSB);
 						midiout.control(chan, 38, valLSB);
+						midiout.control(chan, 101, 127);
+						midiout.control(chan, 100, 127);
 					});
 				};
 


yaxu:
Great thanks !

PR #288, opened 2023-02-18
Create tidal-pitch-glide

fracnesco:
Added one example on the hacks folder to create pitch glide setting a parameter of an active synth

### DIFF
--- a/hacks/tidal-pitch-glide
+++ b/hacks/tidal-pitch-glide
@@ -1,0 +1,60 @@
+/*
+
+How to have pitch glide between notes with Tidal
+
+*/
+
+
+(
+
+var synth;
+var instance = 0;
+
+SynthDef(\glideExample, {|note=0, drive=1, filter=1, glide=0.2, wrate=1, gain=1|
+
+	var freqlag = Lag.kr(note+30, glide);
+	var sig = Lag.ar( VarSaw.ar( freqlag.midicps*[0.99,1.01], [0,0.01*SinOsc.kr(1)]), [0.1,0.11], 100);
+
+	sig = 0.3*atan(sig*10*drive);
+	sig = RLPF.ar(sig, ([1300, 1100, 800, 300]*filter+  SinOsc.ar(wrate).range(100,500)).clip(20,8000), 0.3, [0.2, 0.3, 0.7]);
+   sig = LeakDC.ar(Mix.ar(sig));
+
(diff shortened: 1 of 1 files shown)


hacks/tidal-pitch-glide
```diff
@@ -0,0 +1,60 @@
...
+   sig = LeakDC.ar(Mix.ar(sig));
+
+	Out.ar(0, DirtPan.ar( sig*gain*0.3, ~dirt.numChannels, 0));
+}).add;
+
+
+
+~dirt.soundLibrary.addSynth(\glideExample, (play: {
+
+	if ( (instance ==0),
```
telephon: you could also check directly, `synth.isNil`

hacks/tidal-pitch-glide
```diff
@@ -0,0 +1,60 @@
...
+SynthDef(\glideExample, {|note=0, drive=1, filter=1, glide=0.2, wrate=1, gain=1|
+
+	var freqlag = Lag.kr(note+30, glide);
+	var sig = Lag.ar( VarSaw.ar( freqlag.midicps*[0.99,1.01], [0,0.01*SinOsc.kr(1)]), [0.1,0.11], 100);
+
+	sig = 0.3*atan(sig*10*drive);
+	sig = RLPF.ar(sig, ([1300, 1100, 800, 300]*filter+  SinOsc.ar(wrate).range(100,500)).clip(20,8000), 0.3, [0.2, 0.3, 0.7]);
+   sig = LeakDC.ar(Mix.ar(sig));
+
+	Out.ar(0, DirtPan.ar( sig*gain*0.3, ~dirt.numChannels, 0));
```
telephon: If you use `0` as the output, this SynthDef won't work properly with the rest of superdirt, because it plays right out to your speakers (no routing, etc.).

Probably better you use `out` as an argument to the `Out` UGen.

hacks/tidal-pitch-glide
fracnesco: Thanks!

hacks/tidal-pitch-glide
fracnesco: Yes you are right! Fixed

telephon (commented):
You can leave out the comments, I think, because they just repeat what the code says? Also no extra parentheses needed around `synth.isNil`. I can fix that, but while you are at it :)

Thanks for the contribution!

telephon (approved):

PR #289, opened 2023-04-15
add tanh effect

telephon:
https://scsynth.org/t/our-beloved-tanh/7347


ndr-brt (approved):
nice, I had this in my personal code, great to see it in SuperDirt directly

telephon:
I'm still unsure about the volume adjustment. Also I wonder whether it would be better as a global effect actually.

Bubobubobubobubo:
This would be a great addition.

PR #295, opened 2023-09-15
Change visibility of synth variable in GlobalDirtEffect to public

thgrund:
To use the equalizer `EQui` in the SuperDirtMixer, it's needed to access the `synth` variable of the `GlobalDirtEffect` from outside. This helps to exchange data between the synth and the ui view. I assume this was never needed before, but it should not harm anything.

### DIFF
--- a/classes/GlobalDirtEffect.sc
+++ b/classes/GlobalDirtEffect.sc
@@ -15,7 +15,7 @@ GlobalDirtEffect {
 
 	var <>name, <>paramNames, <>numChannels, <state;
 	var <>alwaysRun = false;
-	var <>synth, defName;
+	var <synth, defName;
 
 	*new { |name, paramNames, numChannels|
 		^super.newCopyArgs(name, paramNames, numChannels, ())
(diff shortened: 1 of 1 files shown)


classes/GlobalDirtEffect.sc
```diff
@@ -15,7 +15,7 @@ GlobalDirtEffect {
 
 	var <>name, <>paramNames, <>numChannels, <state;
 	var <>alwaysRun = false;
-	var synth, defName;
+	var <>synth, defName;
```
telephon: I think a getter would be enough. One shouldn't really try to make a synth from the outside. Or maybe you need this?

telephon (changes requested):

classes/GlobalDirtEffect.sc
thgrund: This is a valid point and I totally agree! I only need the getter and beside that, I create the global effect synth as usual.

telephon (approved):

Issue #302, opened 2024-09-05
Merging StrudelDirt ?


yaxu:
[Strudel](https://strudel.cc) has a kind of superdirt clone [superdough](https://github.com/tidalcycles/strudel/tree/main/packages/superdough) as it's default output. Strudel can also target superdirt via the [osc](https://github.com/tidalcycles/strudel/tree/main/packages/osc) support, which works well but there is some mismatch with the way superdough works, so @daslyfe has recently forked superdirt into [StrudelDirt](https://github.com/daslyfe/StrudelDirt) with some changes.

So I thought I'd make this issue to discuss the possibility of merging them again, with superdirt getting some enhancements in the process. This might involve tweaks on the strudel/tidal side if needed to get things closer.

telephon:
Ah yes, it would be really so much better to keep them together! There are also people who slowly work on a browser version of supercollider, so that may help strudel as well.

Now there are a [lot of changes which are hard to separate](https://github.com/musikinformatik/SuperDirt/pull/303).

We'll need to have a chat about why certain things were needed, it seems that some could be done with less impact.

But please, let's not get superdirt forked away, because that will make development really complicated. @daslyfe , if you can bear with us, please request changes, I am open to them.

daslyfe:
> Ah yes, it would be really so much better to keep them together! There are also people who slowly work on a browser version of supercollider, so that may help strudel as well.
> 
> Now there are a [lot of changes which are hard to separate](https://github.com/musikinformatik/SuperDirt/pull/303).
> 
> We'll need to have a chat about why certain things were needed, it seems that some could be done with less impact.
> 
> But please, let's not get superdirt forked away, because that will make development really complicated. @daslyfe , if you can bear with us, please request changes, I am open to them.

Hi yes, that would be great if a fork is not needed to support these features I am working on that are useful for strudel.

I am sure there are much better ways to implement some of these things, I am new to supercollider and hacking things together at the moment so very open to suggestions! Here are some of the proposed changes: 

Suggested dirt changes: 

Volume envelope adsr instead of asr. Note event 
length (clip/legato) determines the sustain length. Once the note event length is completed the release stage of the envelope begins 


Filter adsr envelopes - Created envelopes for each filter type. Envelope params default to volume adsr params when not specified. 

New oscillators: 
Sawtooth, triangle, sine, square, improved supersaw, pwm with continuous modulation

[...]

PR #303, opened 2024-09-05
Trying to merge back changes from SuperDough fork

telephon:
See https://github.com/musikinformatik/SuperDirt/issues/302

There are lots of changes, so it is a little hard to tell what would impact what. But given this merge request, we can see the diff better.

### DIFF
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -192,7 +192,6 @@ DirtEvent {
 			*[
 				in: orbit.synthBus.index, // read from synth bus, which is reused
 				out: orbit.dryBus.index, // write to orbital dry bus
-				amp: ~amp,
 				gain: ~gain,
 				overgain: ~overgain,
 				sample: ~hash, // required for the cutgroup mechanism
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -144,7 +144,14 @@ DirtEvent {
 	sendSynth { |instrument, args|
 		var group = ~synthGroup;
 		args = args ?? { this.getMsgFunc(instrument).valueEnvir };
+	
 		args.asControlInput.flop.do { |each|
+		// instrument.value.postln;
+		//     if (instrument.value == "supersaw") {
+		// 		"HI".postln;
+		// 	};
+	
+		
 			server.sendMsg(\s_new,
 				instrument,
 				-1, // no id
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -104,8 +104,8 @@ DirtEvent {
 		sustain = ~sustain.value;
 		sustain = sustain ?? {
 			delta = ~delta.value;
-			if(~legato.notNil) {
-				delta * ~legato.value
+			if(~clip.notNil) {
+				(delta / ~cps) * ~clip.value
 			} {
 				unitDuration = unitDuration ? delta;
 				loop !? { unitDuration = unitDuration * loop.abs };
@@ -120,6 +120,7 @@ DirtEvent {
 		~fadeInTime = if(~begin != 0) { ~fadeTime } { 0.0 };
 		if (~timescale.notNil) {sustain = sustain * ~timescale };
 		~sustain = sustain - (~fadeTime + ~fadeInTime);
+		// ~sustain = (~delta.value / ~cps) * ~clip.value ? 1;
 		~speed = speed;
 		~endSpeed = endSpeed;
 
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -113,18 +113,17 @@ DirtEvent {
 				loop !? { unitDuration = unitDuration * loop.abs };
 			}
 		};
-		sustain = sustain + release;
+		// sustain = sustain + release;
 		
 
 		// end samples if sustain exceeds buffer duration
 		// for every buffer, unitDuration is (and should be) defined.
 		if(useUnit) { sustain = min(unitDuration, sustain) };
-
+	
 		~fadeTime = min(~fadeTime.value, sustain * 0.19098);
 		~fadeInTime = if(~begin != 0) { ~fadeTime } { 0.0 };
 		if (~timescale.notNil) {sustain = sustain * ~timescale };
 		~sustain = sustain - (~fadeTime + ~fadeInTime);
-		// ~sustain = (~delta.value / ~cps) * ~clip.value ? 1;
 		~speed = speed;
 		~endSpeed = endSpeed;
 
@@ -170,6 +169,7 @@ DirtEvent {
 				overgain: ~overgain,
 				sample: ~hash, // required for the cutgroup mechanism
 				cut: ~cut.abs,
+				release: ~release,
 				sustain: ~sustain, // after sustain, free all synths and group
 				fadeInTime: ~fadeInTime, // fade in
 				fadeTime: ~fadeTime // fade out
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -13,14 +13,9 @@ DirtEvent {
 		event.parent = orbit.defaultParentEvent;
 		event.use {
 
-		
 			// s and n stand for synth/sample and note/number
 			~s ?? { this.splitName };
 
-		// if (~unison.notNil) {
-		// 		~s = ~s ++ ~unison;		
-		// };
-			
 			// unless orbit wide diversion returns something, we proceed
 			~diversion.(this) ?? {
 				if(~s != \) { // backslash stands for do nothing
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -67,6 +67,7 @@ DirtEvent {
 		var accelerate = ~accelerate.value;
 		var avgSpeed, endSpeed;
 		var useUnit;
+		var release = ~release.value ? 0.0;
 
 		~freq = ~freq.value;
 		unitDuration = ~unitDuration.value;
@@ -102,6 +103,7 @@ DirtEvent {
 		};
 
 		sustain = ~sustain.value;
+	
 		sustain = sustain ?? {
 			delta = ~delta.value;
 			if(~clip.notNil) {
@@ -111,6 +113,8 @@ DirtEvent {
 				loop !? { unitDuration = unitDuration * loop.abs };
 			}
 		};
+		sustain = sustain + release;
+		
 
 		// end samples if sustain exceeds buffer duration
 		// for every buffer, unitDuration is (and should be) defined.
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -122,16 +122,15 @@ DirtEvent {
 	
 		sustain = sustain ?? {
 			delta = ~delta.value;
-			if(~clip.notNil) {
-				(delta / ~cps) * ~clip.value
-			} {
-				unitDuration = unitDuration ? delta;
-				loop !? { unitDuration = unitDuration * loop.abs };
-			}
+			(delta / ~cps) * (~clip.value ? 1)
+			// if(~clip.notNil) {
+			// 	(delta / ~cps) * ~clip.value
+			// } {
+			// 	unitDuration = unitDuration ? delta;
+			// 	loop !? { unitDuration = unitDuration * loop.abs };
+			// }
 		};
-		// sustain = sustain + release;
 		
-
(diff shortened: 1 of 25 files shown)


geikha:
let's also consider that we could extend further the first handshake between Tidal/Strudel to SuperDirt, so that SuperDirt knows which one is connected and maybe interpret differently the messages (majorly on breaking changes)

library/strudel-synths.scd
```diff
@@ -250,20 +250,22 @@
     //emulation of juno 60 chorus
     SynthDef("strudel_chorus" ++ numChannels, { |out, mix, depth, speed|
 		var input = In.ar(out, numChannels);
-        var l;
-        var r;
+        var l = input[0];
```
telephon, line 285: `input` may have any number of channels, and if someone chooses to run it mono, it will not be an array. You could use the technique we use in `dirt_leslie`.

daslyfe: did not know that it was not always an array, Ill fix that!

daslyfe:
> let's also consider that we could extend further the first handshake between Tidal/Strudel to SuperDirt, so that SuperDirt knows which one is connected and maybe interpret differently the messages (majorly on breaking changes)

good callout, hopefully overtime tidal and strudel can be brought closer together in terms of things like note scaling but this is possibly a good short term solution

telephon:
>let's also consider that we could extend further the first handshake between Tidal/Strudel to SuperDirt, so that SuperDirt knows which one is connected and maybe interpret differently the messages (majorly on breaking changes)

What differences would need to be handled? Wouldn't it be much easier to just keep things the same? We are a really small community!

classes/DirtEvent.sc
```diff
@@ -67,6 +83,7 @@ DirtEvent {
 		var accelerate = ~accelerate.value;
 		var avgSpeed, endSpeed;
 		var useUnit;
+		~release = ~release.value ? 0.01;
```
telephon, line 82: default values are usually given in the default event. Unless you need to caclulate something with `~release` here, also the `.value` will be called automatically.

classes/DirtEvent.sc
```diff
@@ -102,24 +119,38 @@ DirtEvent {
 		};
 
 		sustain = ~sustain.value;
+	
 		sustain = sustain ?? {
```
telephon, line 119: The changes below: are you sure strudel couldn't translate its logic to the way tidalcycles works? It will be very confusing to change all that …

classes/DirtEvent.sc
```diff
@@ -129,6 +160,7 @@ DirtEvent {
 		~channel !? { ~pan = ~pan.value + (~channel.value / ~numChannels) };
 		~pan = ~pan * 2 - 1; // convert unipolar (0..1) range into bipolar one (-1...1)
 		~delayAmp = ~delay ? 0.0; // for clarity
+		~z1 = ~z1 ? ~n;
```
telephon, line 159: can you explain what `~z1` is used for?

classes/DirtEventTypes.sc
```diff
@@ -108,15 +109,15 @@ DirtEventTypes {
 
 				if (hasNote) {
 					freq = ~freq.value;
-					~midinote = (freq.cpsmidi).round(1).asInteger;
+					~midinote = ((freq.cpsmidi).round(1) ? 0).asInteger;
```
telephon, line 112: the default is not necessary and can be dropped. `freq.cpsmidi).round(1)` will never be `nil`.

classes/DirtEventTypes.sc
```diff
@@ -108,15 +109,15 @@ DirtEventTypes {
...
 					// Assume aftertouch means no noteOn, for now..
 					if(~polyTouch.notNil) {
 						val = ~polyTouch;
 						note = ~midinote;
 						schedmidi.value({ midiout.polyTouch(chan, note, val) })
 					} {
 						// match dirt_gate SynthDef amplitude scaling
-						~amp =  ~amp.value * pow(~gain.min(2) + ~overgain, 4);
+						~amp =  ~amp.value * StrudelUtils.gainCurve(~gain + ~overgain, 4);
```
telephon, line 120: Let's introduce `gainCurve` to SuperDirt. Should it be global or orbit-wise?

classes/DirtOrbit.sc
```diff
@@ -173,22 +175,24 @@ DirtOrbit {
 			~overgain = 0.0;
 			~cut = 0.0;
 			~unit = \r;
-			~n = \none; // sample number or note
-			~octave = 5;
-			~midinote = #{ ~note ? ~n + (~octave * 12) };
+			~n = nil;
```
telephon, line 185: we should find a way around setting this to nil. Why can't it be `\none`?

telephon: Basically, you can just directly replace the default if you need, no need for StrudelUtils:

```
~dirt.set(\n, 42); // whatever you want
```

classes/DirtOrbit.sc
```diff
@@ -173,22 +175,24 @@ DirtOrbit {
...
 			~lag = 0.0;
 			~length = 1.0;
 			~loop = 1.0;
 			~dry = 0.0;
 			~lock = 0; // if set to 1, syncs delay times with cps
 
-			~amp = 0.4;
+			~amp = 1;
 			~fadeTime = 0.001;
+			~delaytime = 0.1875;
```
telephon, line 201: I am not 100% sure, but I think this means that there is always delay on. This can be set for each piece separately.

classes/DirtSoundLibrary.sc
```diff
@@ -179,7 +203,9 @@ DirtSoundLibrary {
 		};
 
 		files = pathMatch(folderPath.standardizePath +/+ "*"); // dependent on operating system
-		if(sortFiles) { files.sort };
+		if(sortFiles) { 
+			 files.sort;
```
telephon, line 214: in sclang, there is no `;` inside the last statement of a function. It is not wrong, but the change is not needed.

classes/DirtSoundLibrary.sc
```diff
@@ -298,8 +327,7 @@ DirtSoundLibrary {
 			^event
 
 		} {
-			// the index may be \none (a Symbol), but this converts it to 0
-			event = allEvents.wrapAt(index.asInteger);
+			event = allEvents.wrapAt((index ? 0).asInteger);
```
telephon, line 337: This can be revisited in SuperDirt.

classes/DirtSoundLibrary.sc
```diff
@@ -320,8 +348,9 @@ DirtSoundLibrary {
 	}
 
 	makeEventForBuffer { |buffer, metaData|
-		var baseFreq = 60.midicps;
+		var baseFreq = StrudelUtils.baseNote().midicps;
```
telephon, line 358: I intentionally kept separate `DirtSoundLibrary` and `SuperDirt`. This `StrudelUtils.baseNote().` couples both to one system. MAybe we find a way around this?

classes/SuperDirt.sc
```diff
@@ -53,10 +54,14 @@ SuperDirt {
 
 	init {
 		soundLibrary = DirtSoundLibrary(server, numChannels);
+		outputvolume = server.volume;
+		outputvolume.setVolumeRange(-90, 0);
```
telephon: We want to be able to use SuperDirt with other libraries – we shouldn't mess with the server volume, I think.

telephon: I am sure there is another way to set volume? Normally we are in the range [min = -90, max = 6] so here you have found a reason why this should be less?

hacks/jitlib-hacks.scd
```diff
@@ -74,7 +75,7 @@ Pdef(\x,
 		\s, Pn(Pshuf([\bd, \hh, \cp, \imp], 5)),
 		\begin, Pn(Pseries(0, 0.002, 20)),
 		\dur, 0.1,
-		\legato, Pwhite(0.01, 0.4)
+		\clip, Pwhite(0.01, 0.4)
```
telephon, line 78: What is `clip` doing?

library/default-synths-extra.scd
```diff
@@ -220,6 +225,7 @@
 		sound = MoogFF.ar(sound, SinOsc.ar(basefreq/32*rate, 0).range(lfof1,lfof2), resonance*4);
 		sound = MoogFF.ar(sound, min(env2*lfof2*1.1, 22000), 3);
 		sound = sound.tanh*5;
+		sound = sound * StrudelUtils.synthGain;
```
telephon, line 228: No need to hardcode the volume adjustments, you can just do that by setting the `~amp` parameter in the input. So all the below can easily be avoided.

synths/core-modules.scd
```diff
@@ -104,27 +105,27 @@ this may be refacored later.
...
 
-~dirt.addModule('hpf',
-	{ |dirtEvent|
-		dirtEvent.sendSynth("dirt_hpf" ++ ~numChannels,
-			[
-				hcutoff: ~hcutoff,
-				hresonance: ~hresonance,
-				out: ~out
-		])
+// ~dirt.addModule('hpf',
```
telephon, line 108: why do you remove `hpf` and `lpf` – are there problems with them?

synths/core-synths-global.scd
```diff
@@ -71,62 +74,45 @@ CORE SYNTHDEFS FOR DIRT
 	dryBus and effectBus have \ir as rates (they shouldn't change at runtime)
 
 	*/
-
-
-
-	if(\SwitchDelay.asClass.notNil) {
-
-		SynthDef("dirt_delay" ++ numChannels, { |dryBus, effectBus, gate = 1, delaytime, delayfeedback, delaySend = 1, delayAmp = 1, lock = 0, cps = 1|
+	SynthDef("dirt_shimmer" ++ numChannels, { |dryBus, effectBus, gate = 1, delaytime, delayfeedback, shimmer = 0, lock = 1, cps = 1|
```
telephon, line 77: let's move shimmer to SuperDirt ...

synths/core-synths.scd
```diff
@@ -193,9 +206,9 @@ live coding them requires that you have your SuperDirt instance in an environmen
 
 	*/
 
-	SynthDef("dirt_out" ++ numChannels, { |out, to, sustain = 1, fadeInTime = 0.001, fadeTime = 0.001, through = 0, amp = 1|
+	SynthDef("dirt_out" ++ numChannels, { |out, to, totalDuration, fadeInTime = 0.001, fadeTime = 0.001, through = 0, amp = 1|
 		var signal = In.ar(out, numChannels);
-		var env = EnvGen.ar(Env([0, 1, 1, 0], [fadeInTime, sustain, fadeTime], \sin));
+		var env = EnvGen.ar(Env([0, 1, 1, 0], [fadeInTime, totalDuration - (fadeInTime + fadeTime), fadeTime], \sin));
```
telephon, line 211: yes, in supercollider generally, `sustain` means `totalDuration` – this is a bit unusual, but the standard …

telephon:
@daslyfe I've left a lot of comments for you so it hopefully makes it easier to separate those things that are really needed and those that just happened due to a lack of documentation in superdirt.

In the meantime, since superdirt is live hackable, many of the changes could just be moved to a file that is evaluated at startup. Like all the SynthDefs etc.

Maybe you could check and see if you can separate things a little? That would be immensly helpful.

daslyfe:
> @daslyfe I've left a lot of comments for you so it hopefully makes it easier to separate those things that are really needed and those that just happened due to a lack of documentation in superdirt.
> 
> In the meantime, since superdirt is live hackable, many of the changes could just be moved to a file that is evaluated at startup. Like all the SynthDefs etc.
> 
> Maybe you could check and see if you can separate things a little? That would be immensly helpful.

Thanks for taking the time to comb through all of this! It helps to know how things can be better separated. I am still in a “hacking stuff together” phase of making superdirt strudel compatible. I intend to break changes into smaller much more manageable PRs

telephon:
that is good! You could take my comments as things to mind to be able to do it with less effort / changes.

telephon:
@daslyfe I just saw that you changed the monitor and added wet and dry. https://github.com/musikinformatik/SuperDirt/blob/ad53027ef969009433725ca9fc8ce8bfc20554e7/classes/DirtOrbit.sc#L66 and https://github.com/musikinformatik/SuperDirt/blob/ad53027ef969009433725ca9fc8ce8bfc20554e7/synths/core-synths-global.scd#L26

Is there a reason why the local wet / dry controls of each effect are not sufficient?

PR #305, opened 2024-09-15
Active / deactive global effects

### LINKED ISSUE #299: dirt_rms ambiguity. should it be running to not?
a minor issue...

in DirtOrbit the global effect dirt_rms is loaded and set as .alwaysRun_(true).
but initially it won't run because of a missing default for the rmsReplyRate argument.

the missing argument is found in core-synths-global.scd...
```SynthDef("dirt_rms" ++ numChannels, { |gate = 1, dryBus, effectBus, rmsReplyRate, rmsPeakLag, orbitIndex|```

changing it to...
```SynthDef("dirt_rms" ++ numChannels, { |gate = 1, dryBus, effectBus, rmsReplyRate = 20, rmsPeakLag = 3, orbitIndex|```
would make the dirt_rms module run as expected.

thoughts?

### PR DESCRIPTION
**Pull request by telephon**
Allow us to activate and deactivate global effects.

This fixes #299.

### DIFF
--- a/classes/GlobalDirtEffect.sc
+++ b/classes/GlobalDirtEffect.sc
@@ -67,8 +67,10 @@ GlobalDirtEffect {
 	}
 
 	resume {
-		synth.run;
-		synth.set(\resumed, 1)
+		synth.server.sendBundle(nil,
+			['/n_run', synth.nodeID, 1]
+			['/n_set', 'resumed', 1]
+		);
 	}
 
 	active_ { |flag|
--- a/classes/GlobalDirtEffect.sc
+++ b/classes/GlobalDirtEffect.sc
@@ -14,7 +14,7 @@ sends only OSC when an update is necessary
 GlobalDirtEffect {
 
 	var <>name, <>paramNames, <>numChannels, <state;
-	var <>alwaysRun = false;
+	var <>alwaysRun = false, <active = true;
 	var <synth, defName;
 
 	*new { |name, paramNames, numChannels|
@@ -23,11 +23,13 @@ GlobalDirtEffect {
 
 	play { |group, outBus, dryBus, effectBus, orbitIndex|
 		this.release;
-		synth = Synth.newPaused(name.asString ++ numChannels,
-			[\outBus, outBus, \dryBus, dryBus, \effectBus, effectBus, \orbitIndex, orbitIndex] ++ state.asPairs,
-			group,
-			\addAfter
-		)
+		if(active) {
+			synth = Synth.newPaused(name.asString ++ numChannels,
+				[\outBus, outBus, \dryBus, dryBus, \effectBus, effectBus, \orbitIndex, orbitIndex] ++ state.asPairs,
+				group,
+				\addAfter
+			)
+		};
 	}
 
 
@@ -52,6 +54,7 @@ GlobalDirtEffect {
 				state[key] = value;
 			}
 		};
+		"someArgsNotNil: %".format(someArgsNotNil, event).postln;
 		if(someArgsNotNil) { this.resume };
 		if(argsChanged.notNil) {
 			synth.set(*argsChanged);
@@ -59,8 +62,15 @@ GlobalDirtEffect {
 	}
 
 	resume {
-		synth.run;
-		synth.set(\resumed, 1)
+		if(active) {
+			synth.run;
+			synth.set(\resumed, 1)
+		}
+	}
+
+	active_ { |flag|
+		active = flag;
+		if(active) { this.resume }
 	}
 
 	printOn { |stream|
--- a/classes/GlobalDirtEffect.sc
+++ b/classes/GlobalDirtEffect.sc
@@ -42,19 +42,23 @@ GlobalDirtEffect {
 		};
 	}
 
+	// set is called for each event, If paused and active and parameters changed, it resumes the synth.
 	set { |event|
 		var argsChanged, someArgsNotNil = alwaysRun;
-		paramNames.do { |key|
-			var value = event[key];
-			value !? { someArgsNotNil = true };
-			if(state[key] != value) {
-				argsChanged = argsChanged.add(key).add(value);
-				state[key] = value;
+
+		if(active) {
+			paramNames.do { |key|
+				var value = event[key];
+				value !? { someArgsNotNil = true };
+				if(state[key] != value) {
+					argsChanged = argsChanged.add(key).add(value);
+					state[key] = value;
+				}
+			};
+			if(someArgsNotNil) { this.resume };
+			if(argsChanged.notNil and: { synth.notNil }) {
+				synth.set(*argsChanged);
 			}
-		};
-		if(someArgsNotNil) { this.resume };
-		if(argsChanged.notNil and: { synth.notNil }) {
-			synth.set(*argsChanged);
 		}
 	}
 
@@ -63,10 +67,8 @@ GlobalDirtEffect {
 	}
 
 	resume {
-		if(active) {
-			synth.run;
-			synth.set(\resumed, 1)
-		}
+		synth.run;
+		synth.set(\resumed, 1)
 	}
 
 	active_ { |flag|
--- a/classes/GlobalDirtEffect.sc
+++ b/classes/GlobalDirtEffect.sc
@@ -23,13 +23,11 @@ GlobalDirtEffect {
 
 	play { |group, outBus, dryBus, effectBus, orbitIndex|
 		this.release;
-		if(active) {
-			synth = Synth.newPaused(name.asString ++ numChannels,
-				[\outBus, outBus, \dryBus, dryBus, \effectBus, effectBus, \orbitIndex, orbitIndex] ++ state.asPairs,
-				group,
-				\addAfter
-			)
-		};
+		synth = Synth.newPaused(name.asString ++ numChannels,
+			[\outBus, outBus, \dryBus, dryBus, \effectBus, effectBus, \orbitIndex, orbitIndex] ++ state.asPairs,
+			group,
+			\addAfter
+		)
 	}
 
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: classes/GlobalDirtEffect.sc [hunk 1]
```diff
@@ -23,11 +23,13 @@ GlobalDirtEffect {
 
 	play { |group, outBus, dryBus, effectBus, orbitIndex|
 		this.release;
-		synth = Synth.newPaused(name.asString ++ numChannels,
-			[\outBus, outBus, \dryBus, dryBus, \effectBus, effectBus, \orbitIndex, orbitIndex] ++ state.asPairs,
-			group,
-			\addAfter
-		)
+		if(active) {
```
telephon: actually, this may not be needed. It is ok, to have a paused synth in there. I'll fix this.

telephon:
OK, maybe you can check this?

thgrund:
> OK, maybe you can check this?

I am eager to help :-) 
More details here: https://github.com/musikinformatik/SuperDirt/pull/307#issuecomment-2383967186

But we can continue the discussion in this PR.

thgrund:
Hey @telephon ! So in general the changes are doing what I expect. But there is one issue, that harte_echtzeit and me faced during our tests.

It looks like, that there is a race condition between the evaluation of the `active` and the `set` function.  This can be reproduced with a simple example: 

Single note example in TidalCycles
```
  d1 $ s "superpiano" <| note "c" # room 0.8 # sz 0.8
 ```

When you activate and deactivate the reverb effect in SuperCollider in between the events, then everything works flawless: 
```
 ~dirt.orbits[0].globalEffects[4].active_(true) -- assuming the reverb effect is on the fifth position
~dirt.orbits[0].globalEffects[4].active_(false) -- assuming the reverb effect is on the fifth position
```

But when you try the same with i.e. more notes (or simply when more events will be send), then sometimes the effect will be resumed immediately, even when the active property is false. Like in this example: 

```
  d1 $ s "superpiano" <| note "c*32" # room 0.8 # sz 0.8
```

**Comment by thgrund** [comment]
@telephon can I help here to speed up this Pull Request? As in my [Pull Request](https://github.com/musikinformatik/SuperDirt/pull/307#issuecomment-2383967186), I would recommend to not execute `set` if the global effect is not active. If you don't want to do that directly inside of `GlobalDirtEffect.sc`, then you could check the flag when `set` is called inside of `setGlobalEffects` in `DirtOrbit`.

WDYT?

**Comment by telephon** [comment]
Hi @thgrund I'll return to this next week – thank you for your testing and of course also the reminder.
For now, one question:

The race condition happens when you are toggling the effects while you play notes from tidal, right?

**Comment by thgrund** [comment]
Hey @telephon thanks for letting me know :-) 

> The race condition happens when you are toggling the effects while you play notes from tidal, right?

Exactly!

**Comment by telephon** [comment]
I've just made a change, could you check if it fixes your issue?

**Comment by thgrund** [comment]
I finished my tests and I am a little bit confused now. The race condition still exists. I do not fully understand why this happens, but from the clients code perspective this looks good. I tried it with my compressor effect and with the SuperDirt reverb effect. 

Here is an example with some debug messages, where I logged the function calls in `GlobalDirtEffect.sc`: 

```
set; active: true, fxName: dirt_global_compressor
resume; active: true, fxName: dirt_global_compressor
set; active: true, fxName: dirt_global_compressor
resume; active: true, fxName: dirt_global_compressor
active_; flag: false, fxName: dirt_global_compressor
pause; active: false, fxName: dirt_global_compressor, synth: Synth('dirt_global_compressor2' : 1200)
set; active: false, fxName: dirt_global_compressor
set; active: false, fxName: dirt_global_compressor
set; active: false, fxName: dirt_global_compressor
set; active: false, fxName: dirt_global_compressor
set; active: false, fxName: dirt_global_compressor
```

But the effect was not bypassed reliable when I send TidalCycles events. Another interesting aspect is, that when I changed the `set` function like this: 

```
          set { |event|
                  var argsChanged, someArgsNotNil = alwaysRun;
                  if(active) {
                          paramNames.do { |key|
                                  var value = event[key];
                                  value !? { someArgsNotNil = true };
                                  if(state[key] != value) {
                                          argsChanged = argsChanged.add(key).add(value);
                                          state[key] = value;
                                  }
                          };
                          if(someArgsNotNil) { this.resume };
                          if(argsChanged.notNil and: { synth.notNil }) {
                                  synth.set(*argsChanged);
                          }
                  } {
                      this.pause;
                  }
```

With this changes, I hear the synth bypassed when I call `active_(false)` on the effect. After that it becomes active again for a short amount of time automatically and then it finally gets paused. Could it be, that the pause action on the server side takes some time on more complex SynthDefs and when this happens, then the client code is not applied like we expect it?

**Comment by telephon** [comment]
> Could it be, that the pause action on the server side takes some time on more complex SynthDefs and when this happens, then the client code is not applied like we expect it?

no, I don't think so, pausing is very fast. The mistake is probably in the interaction of the three following things:
1. resuming of self-paused synths (global effect synths pause themselves when not used)
2. resuming of explictly paused synths
3. pausing of running synths and setting the active flag= true

**Comment by telephon** [comment]
Just to reduce the possibilities of wrong ordering, I've added a little change.

**Comment by thgrund** [comment]
Hey @telephon , I faced an issue with the changes you made in the `resume` function. To solve this, I changed it locally in this way: 

```
resume {
	synth.server.sendBundle(nil,
		['/n_run', synth.nodeID, 1],
		['/n_set', synth.nodeID, 'resumed', 1]
	);
}
``` 

For the analysis it would be helpful to know, when the synth is really paused. Is there a way to achieve that? I found the `NodeWatcher` class in the SuperCollider documentation, but it didn't worked for me in this case. Or to be more precise, when I check the synth whether it is playing or running, then the value is always false. I will keep trying it.

I will do my investigation in parallel and will inform you, when I find something.

**Comment by telephon** [comment]
Thanks for the catch, I am sorry I didn't check carefully enough.

Also, I've added a change so that the synth is registered. You should be able to check if it is running or not (there is, of course a tiny delay between the state change in scsynth and the state change in sclang).

```supercollider
a = ~dirt.orbits[0].globalEffects[0];
a.synth.isPlaying; // exists or not
a.synth.isRunning; // paused or not
```

telephon:
> Or to be more precise, when I check the synth whether it is playing or running, then the value is always false. I will keep trying it.

Yes, this is the case when you don't register them explicitly.

thgrund:
Hey @telephon . Here are some test results for one effect for one orbit where you can see, that sometimes the synth will be paused, and sometimes it's still running. What I learned from these examples is, that the action calls from the client sides are exactly the same (active, pause, set). But the result differs. At this point I would expect, that the error happens on the server side. 

My next step is to print the OSc messages that are send to the server and printing the polls from the DirtPause UGen.

This is the success case:
```
// active = true (works)
active; flag: true; synth status: true
resumed; synth status: true

// usual set for active synth (works)
before set; synth status: true; active: true
resumed; synth status: true
set; synth status: true

// active = false
active; flag: false; synth status: true
pause; synth status: true

// set call - no calculation (first one may be caused by server response delay)
before set; synth status: true; active: false
before set; synth status: false; active: false
before set; synth status: false; active: false
before set; synth status: false; active: false
before set; synth status: false; active: false
```

Error case (synth keeps running):
```
// active = true (works) - synth status = false -> server delay
active; flag: true; synth status: false
resumed; synth status: false

// usual set for active synth (works)
before set; synth status: true; active: true
resumed; synth status: true
set; synth status: true

// active = false
active; flag: false; synth status: true
pause; synth status: true

//*******************************
// HERE THE ERROR HAPPENS
//*******************************
//
// set call - no calculation; but synth runs and active = false
before set; synth status: true; active: false
before set; synth status: true; active: false
before set; synth status: true; active: false
before set; synth status: true; active: false
```

telephon:
This is excellent, thank you for the good analysis. Just one question, what is meant by "synth status: true"? That it is playing (exists on the server), but may be paused (then active=false)?

thgrund:
Ah sorry, this was not clear. The synth status is the value of `isRunning`. Because the Node always exists, means `isPlaying` is always true. This is why I just printed the value of `synth.isRunning` for one effect. Basically like this: 
```
format("resumed; synth status: %", synth.isRunning).postln;}
```

telephon:
So yes, this is clearly a timing issue, the synth is woken up after it is paused explicitl by a message that is already on its way.
I have done two things: use an existing pauseImmediately parameter and add the standard latency to the osc messages, so they come late enough.

Could you please check again? Sorry, this is a little tricky.

classes/GlobalDirtEffect.sc
```diff
@@ -64,13 +64,16 @@ GlobalDirtEffect {
...
-		if(synth.notNil) { synth.run(false) }
+		synth.server.sendBundle(nil,
+			['/n_set', synth.nodeID, 'pauseImmediately', 1],
+			['/n_run', synth.nodeID, 0]
+		);
 	}
 
 	resume {
 		synth.server.sendBundle(nil,
 			['/n_run', synth.nodeID, 1]
```
thgrund: Here is a missing comma.

classes/GlobalDirtEffect.sc
```diff
@@ -64,13 +64,16 @@ GlobalDirtEffect {
 	}
 
 	pause {
-		if(synth.notNil) { synth.run(false) }
+		synth.server.sendBundle(nil,
```
thgrund: We need to add the condition again (`if(synth.notNil) {...}`)

thgrund (changes requested):
Nice! Adding the synth server latency resolves the issue! 🥳

classes/SuperDirtUGens.sc
```diff
@@ -189,9 +189,10 @@ DirtPause {
 	*ar { | signal, graceTime = 1, pauseImmediately = 0 |
 		// immediately pause when started
 		PauseSelf.kr(Impulse.kr(0) * pauseImmediately);
+		"DirtPause % running".format(UGen.buildSynthDef.name).poll(2);
```
thgrund: And I think this line can be removed now?

telephon:
if all works, I'd merge into develop now.

thgrund:
I just checked it again a few seconds ago and from my end everything works fine now :-)

PR #307, opened 2024-09-26
Add bypass property for global dirt effects

thgrund:
I want to bypass global effects and already found this MR (https://github.com/musikinformatik/SuperDirt/pull/305). I implemented a similar mechanism, but I just wanted to bypass the global effect and pause the synth. When I compare the codes, then the behaviour differs, I assume. Or would it be possible to use the other implementation to achieve a similare result? The major difference that I see is, that in my MR I pause the synth and when `set` is called, then nothing happens. But the global effect stays in the node tree and inside of the global effect array.

Is it worth to combine these two MRs?

Example how I use my custom bypass mechanism successfully in the SuperDirtMixer:
- https://github.com/thgrund/SuperDirtMixer/blob/mrreason-setup/classes/uiComponents/CompressorUI.sc#L106-L110

My use case is to bypass the compressor synth and the equalizer synth in my rework of the SuperDirtMixer. The current version can be found here (a screenshot can be found inside of the README.md):
- https://github.com/thgrund/SuperDirtMixer/tree/mrreason-setup

### DIFF
--- a/classes/GlobalDirtEffect.sc
+++ b/classes/GlobalDirtEffect.sc
@@ -14,7 +14,7 @@ sends only OSC when an update is necessary
 GlobalDirtEffect {
 
 	var <>name, <>paramNames, <>numChannels, <state;
-	var <>alwaysRun = false;
+	var <>alwaysRun = false, <>bypass = false;
 	var <synth, defName;
 
 	*new { |name, paramNames, numChannels|
@@ -23,6 +23,7 @@ GlobalDirtEffect {
 
 	play { |group, outBus, dryBus, effectBus, orbitIndex|
 		this.release;
+
 		synth = Synth.newPaused(name.asString ++ numChannels,
 			[\outBus, outBus, \dryBus, dryBus, \effectBus, effectBus, \orbitIndex, orbitIndex] ++ state.asPairs,
 			group,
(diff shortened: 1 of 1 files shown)


telephon:
> When I compare the codes, then the behaviour differs, I assume. Or would it be possible to use the other implementation to achieve a similare result?

I think my recent implementation should achieve the same effect: deactivate the global effect. So it remains paused until activated again, then it bahaves as usual (self-pausing when not used, but activated when set).

Maybe you can check if this is what you need?

thgrund:
Yes, indeed. That's what I need. Gosh...next time I should take a look at the open pull requests. Then I would save some time 👀 But what a coincidence that this topic was opened again when I needed it. I close this PR and I am looking forward to see the changes of #305 in the develop branch. 

I will wait until the PR #305 is published, before I release the SuperDirtMixer rework.

telephon:
Well, this has happened to me also some times. Did you test the branch? I would like to publish soon, but I'd need someone to just run it a bit against their system.

thgrund:
I checked the branch out and tested it locally, once. Just to see whether it behaves in the SuperDirtMixer as I expect it. 

> but I'd need someone to just run it a bit against their system.

I am eager to help :-) On Wednesday I have a beta test of my SuperDirtMixer and will test the branch with someone else. And of course I will run the feature branch against my system and will try out some of my "normal" and extended use cases.

I will share the results with you on Thursday.

Issue #231, opened 2021-02-09
Adding global effect examples in hack folder doesn't work properly


fracnesco:
The example doesn't work properly, for a reference on how to add global effect see
https://github.com/ndr-brt/sc-adente/blob/master/synths/mutable.scd#L227 
from line 227
https://github.com/ndr-brt/sc-adente/blob/master/superdirt_startup.scd#L31
from line 31 to 39

telephon:
Thank you, I'll take a look in a little while!

PR #312, opened 2025-03-31
sound library banks: add an underscore between name and bank, like in strudel.

telephon:
As discussed here: #231
It may break code, but since the feature was unknown, and easy to fix, it should be good.

I've added a convenience method, you can now write: 
```supercollider
~dirt.loadSoundFilesToBank(bankName: "fufu");
```

Then in tidal:
```haskell
s "bd sd rd" # bank "fufu"
```

or synonymously:

```haskell
s "fufu_bd fufu_sd fufu_rd" 
```

### DIFF
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -10,8 +10,7 @@ DirtEvent {
 	play {
 		event.parent = orbit.defaultParentEvent;
 		event.use {
-			// s and n stand for synth/sample and note/number
-			~s ?? { this.splitName };
+			this.splitName;
 			// unless orbit wide diversion returns something, we proceed
 			~diversion.(this) ?? {
 				if(~s != \) { // backslash stands for do nothing
@@ -34,14 +33,19 @@ DirtEvent {
 		orbit.server.makeBundle(~latency, func)
 	}
 
+	// s and n stand for synth/sample and note/number
 	splitName {
-		var s, n;
-		#s, n = ~sound.asString.split($:);
-		if(~bank.notNil) {
-			s = format("%_%", ~bank, s)
+		var sound, note;
+		if(~s.isNil) {
+			#sound, note = ~sound.asString.split($:);
+		} {
+			sound = ~s;
 		};
-		~s = s.asSymbol;
-		~n = if(n.notNil) { n.asFloat } { 0.0 };
+		if(~bank.notNil) {
+			sound = format("%_%", ~bank, sound)
+		};
+		~s = sound.asSymbol;
+		~n = if(note.notNil) { note.asFloat } { 0.0 };
 	}
 
 	mergeSoundEvent {
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -37,7 +37,9 @@ DirtEvent {
 	splitName {
 		var s, n;
 		#s, n = ~sound.asString.split($:);
-		if(~bank.notNil) { s = ~bank ++ s };
+		if(~bank.notNil) {
+			s = format("%_%", ~bank, s)
+		};
 		~s = s.asSymbol;
 		~n = if(n.notNil) { n.asFloat } { 0.0 };
 	}
--- a/classes/DirtEvent.sc
+++ b/classes/DirtEvent.sc
@@ -45,7 +45,7 @@ DirtEvent {
 			sound = format("%_%", ~bank, sound)
 		};
 		~s = sound.asSymbol;
-		~n = if(note.notNil) { note.asFloat } { 0.0 };
+		~n = if(note.notNil) { note.asFloat } { ~n };
 	}
 
 	mergeSoundEvent {
--- a/classes/DirtSoundLibrary.sc
+++ b/classes/DirtSoundLibrary.sc
@@ -147,9 +147,10 @@ DirtSoundLibrary {
 		};
 	}
 
-	// a bank is defined simply by a naming convention
 	loadSoundFilesToBank { |paths, appendToExisting = false, bankName|
-		this.loadSoundFiles(paths, appendToExisting, { |path| bankName ++ "_" ++ path.basename })
+		var form = "%_%";
+		var namingFunc = if(bankName.isNil) { Error("to load into a bank, you have to give a bank name").throw };
+		this.loadSoundFiles(paths, appendToExisting, { |path| format(form, bankName, path.basename) })
 	}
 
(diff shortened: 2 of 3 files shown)


yaxu:
@geikha 's [tidal-drum-machines project](https://github.com/geikha/tidal-drum-machines/) uses `-` as the delimiter in the filesystem.

Strudel [switches this to `_` in the mapping](https://github.com/tidalcycles/strudel/blob/9b4574c7b0dac3627e59207adfc560570b7a7885/website/public/tidal-drum-machines.json#L4), probably because `-` isn't a  valid identifier char in the mininotation

I'm not sure how best to fix this.. Could do it in the filenames in the repo but would have to fix strudel at the same time.  Could also just duplicate the repo.. Otherwise SuperDirt could read the same json file format.

(It also [defines some bank shorthands](https://github.com/tidalcycles/strudel/blob/9b4574c7b0dac3627e59207adfc560570b7a7885/website/public/tidal-drum-machines-alias.json) so e.g. `RolandTR909` can be specified just as `tr909`)

classes/DirtEvent.sc
```diff
@@ -37,7 +37,9 @@ DirtEvent {
 	splitName {
 		var s, n;
 		#s, n = ~sound.asString.split($:);
-		if(~bank.notNil) { s = ~bank ++ s };
+		if(~bank.notNil) {
```
yaxu, line 44: I think splitName isn't called if `s` is defined, and tidal sends `s` rather than `sound`, so this isn't reached.

telephon: ah that's true. I'll have to do it a little difefrently then.

telephon:
Can you test this one?

telephon:
> Otherwise SuperDirt could read the same json file format.

Yes, sure, it could. But unless there is a really clear standard, this just adds complexity ...

telephon:
I can add a delimiter argument to `loadSoundFilesToBank`, or just ask the user to append the delimiter to the name:

```
~dirt.loadSoundFilesToBank(bankName: "fufu_");
```

yaxu:
Yes that works for me thanks!

I think it is probably easier to just rearrange the samples rather than make a level of indirection for now. It seems symlinks work so we could do it that way.

telephon:
> I think it is probably easier to just rearrange the samples rather than make a level of indirection for now. It seems symlinks work so we could do it that way.

How would you have tio change them?

yaxu:
Mainly just change `-` for  `_`.  https://github.com/geikha/tidal-drum-machines/tree/main/machines/RolandTR909

The folder names are also quite long, e.g. `tr909` would be better in this case.

If  a `namingFunction` was passed through to `loadSoundFiles`, then that would allow a lot of flexibility, something like this:

```
~dirt.loadSoundFilesToBank('./machines/RolandTR909', bankName: "tr909", namingFunction: {arg x; x.split('-').pop()});
```

Then we could just add a `.scd` file to repos like `tidal-drum-machines` that can be included in startups to get all the samples. I'm not sure how to express a sample path that would work inside the .scd file though.. with `nowExecutingPath`?

telephon:
Also, this has slipped under the radar, I'm sorry. What si the status @yaxu, we should still do it as you suggested?

geikha:
Woah just saw this too! I agree the names are long, I had done a survey back then and people majorly voted for full names, that's why I had them like that. To me the best case scenario would be developing some sort of info file that we can add to folders for aliases and other stuff

telephon:
> To me the best case scenario would be developing some sort of info file that we can add to folders for aliases and other stuff

That would be easy once we know what we want.

telephon:
> If a `namingFunction` was passed through to `loadSoundFiles`, then that would allow a lot of flexibility, something like this:
> 
> ```
> ~dirt.loadSoundFilesToBank('./machines/RolandTR909', bankName: "tr909", namingFunction: {arg x; x.split('-').pop()});
> ```

I've made a change so the namingFunction is now passed through. What is not so practical right now is that if you pass it in you have to already know the bank name. We could pass the name to the function to curry out another function to pass on.

One question, when you wrote: `{ arg x; x.split('-').pop() }` this would require dashes in all names. I suppose you wanted something that always works, right?


> Then we could just add a `.scd` file to repos like `tidal-drum-machines` that can be included in startups to get all the samples. I'm not sure how to express a sample path that would work inside the .scd file though.. with `nowExecutingPath`?

Yes, almost. We can should `resolveRelative` to resolve the path relative to the file we are in.

geikha:
The original drum machine repo uses `{|x| x.basename.replace("-","")}` as its `namingFunction`. Btw, since I'm clarifying some of the original desicions, the reason I added a `-` in between all drum machines was to make it easy to rename banks or elements (I can imagine someone downloading the repo and changing all `sd` to `sn`).

---

I'd imagine the info file having the following (pseudocode ahead, i'll use the TR-808 as an example):
```
name: "rolandtr808"
type: bank // could be either "bank" or "folder"
aliases: ["tr808", "808"]
ignore: false // allows you to tell SD to ignore the bank/folder without deleting it
```
For folders maybe we could have explicit orders. Also wondering how such a system could handle the `sd` vs `sn` scenario I mentioned above without manually adding an alias to all the `sd` folders.

classes/DirtEvent.sc
```diff
@@ -34,12 +33,19 @@ DirtEvent {
...
+		if(~s.isNil) {
+			#sound, note = ~sound.asString.split($:);
+		} {
+			sound = ~s;
+		};
+		if(~bank.notNil) {
+			sound = format("%_%", ~bank, sound)
+		};
+		~s = sound.asSymbol;
+		~n = if(note.notNil) { note.asFloat } { 0.0 };
```
bsssssss: While testing this i noticed sample selection was broken (seems like it always load `sample:0.0` regardless the value of `~n`).
I think `~n = if(note.notNil) { note.asFloat } { ~n };` fixes it ?

telephon: oh yes, of course (a typo). Thanks!

az4g:
Any chance for this PR to be merged someday? I added it to my local install, it works like a charm and could be useful to other users.

telephon:
Ah it works? So this isn't a problem?: https://github.com/musikinformatik/SuperDirt/pull/312/files#r2021736182

telephon:
OK, I understood, yes it is fixed.

az4g:
Thanks!

Issue #304, opened 2024-09-05
Original dirty compressor


yaxu:
One thing missing from SuperDirt is the original compressor.

https://github.com/tidalcycles/Dirt/blob/071fd88b3e004a11215afd11b718e92d3ab44d18/audio.c#L770-L777

Thankfully [Pulu has ported it](https://club.tidalcycles.org/t/original-dirt-compressor-for-superdirt/5423), code below. It sounds excellent, very useful for from-scratch performance. I think it would be great to incorporate it into superdirt, although probably default off. It needs miSCellaneous_lib as a dependency, perhaps it's better to hard fork [Fb1](https://github.com/dkmayer/miSCellaneous_lib/blob/9c17f5565bb8c8afebdd293a6b0ed697a561aae6/Classes/Nonlinear/Fb1.sc)?

```
// emulation of master compressor from original Dirt
// requires Fb1 from miSCellaeous for single-sample feedback
// https://github.com/dkmayer/miSCellaneous_lib
(
Ndef(\dirtComp, { |amount=1, speed=50|
	var in, max, env;
	in = In.ar(0, ~dirt.numChannels);

	max = Select.ar(ArrayMax.ar(in.abs)[1], in);
	env = Fb1({ |in, out|
		var env, prod;
		env = out[1];
		env = env + (speed / SampleRate.ir);
		prod = (in[0] * env).abs;
		Select.kr(prod > 1, [env, env/prod]);
	}, max, leakDC: false);

	ReplaceOut.ar(0, in * amount.linlin(0, 1, 1, env * 0.2));
}).play(
	group: RootNode(Server.default),
	addAction: \addToTail
);
)

// can also tune the parameters
Ndef(\dirtComp).set(\amount, 1);
Ndef(\dirtComp).set(\speed, 200);
```

telephon:
The UGen is a control rate ugen (Select.kr) so it doesn't need Fb1 really.

Can't test it right now, but this should be about what is needed:

```supercollider

(
Ndef(\dirtComp, { |amount=1, speed=50|
	var input, max, env, n, prod;
	var ctrl;
	n =  2; //~dirt.numChannels;
	input = In.ar(0, n);
	ctrl = input[0];// maybe we want to multichannel expand or sum rathen than taking the first channel?

	max = ArrayMax.ar(input.abs);
	env = LocalIn.kr(n);
	env = env + (speed / SampleRate.ir); // needs fix: making it depend on the sample rate and on block size 
[...]

PR #314, opened 2025-04-22
Add compressor

thgrund:
Topic discussion happens in this issue: 
https://github.com/musikinformatik/SuperDirt/issues/304

My focus with this compressor implementation was to establish a lightweight effect for livecoding environments. It comes with a limited set of parameters. I.e. I added a makeup gain logic instead of setting this makeup gain manually. 
The foundation of this compressor is my compressor implementation in the SuperDirtMixer (https://github.com/thgrund/SuperDirtMixer/blob/main/synths/synth.scd#L24-L67). The compressor implementation in the SuperDirtMixer has more possibilities to tweak the parameters. 

The main parameter for this is `cThresh`. A value of one means no compression and a value close to 0 high compression (threshold is the amplitude value: 1 amp = 0 dB and 0.001 amp = -60 dB) 

You can test it with these events: 

Extreme settings:
```
(type:\dirt, dirt: ~dirt, s: \superpiano, note: 0, orbit: 0, cThresh: 0.001, gain: 0.75, attack: 0.01, sustain: 0.9, release: 0.8).play;
```

Adjusted ratio:

```
(type:\dirt, dirt: ~dirt, s: \superpiano, note: 0, orbit: 0, cThresh: 0.001, cRatio: 2, gain: 0.75, attack: 0.01, sustain: 0.9, release: 0.8).play;
```

No compression at all: 
```
(type:\dirt, dirt: ~dirt, s: \superpiano, note: 0, orbit: 0, attack: 0.01, sustain: 0.9, release: 0.8).play;
```

All these parameters could set from the outside, but are static yet: 
https://github.com/musikinformatik/SuperDirt/pull/314/files#diff-09ee35f9f1ac58413eab41c2ea003e41c624d845d7fdf1e1b9625836aabd59faR438-R442

It's just a design decision, but everytime the signal will clip, I apply a softclip: 
https://github.com/musikinformatik/SuperDirt/pull/314/files#diff-09ee35f9f1ac58413eab41c2ea003e41c624d845d7fdf1e1b9625836aabd59faR471-R475


telephon:
For some reason I never saw this note, @thgrund sorry!

library/default-effects-extra.scd
```diff
@@ -412,5 +412,70 @@ DEFAULT EFFECTS EXTRA
...
+
+    	cLookaheadDelay = DelayC.ar(in, 0.5, (cAttack + cRelease + cKneelag).lag(0.4));
+
+        compressedSound = cLookaheadDelay * gainReduction;
+
+        makeupGain = (cThresh.ampdb * (1 - (1 / cRatio))).neg.dbamp;
+        in = compressedSound * makeupGain;
+
+        // Do a softclip, when clipping
+    	in = (in + K2A.ar(0)).softclip;
```
telephon: This should not be necessary: `in + K2A.ar(0)`. What made you write it?

library/default-effects-extra.scd
```diff
@@ -412,5 +412,70 @@ DEFAULT EFFECTS EXTRA
 			hpfCutoffFreq
 		);
 		ReplaceOut.ar(out, signal)
-}, [\ir]).add;
+    }, [\ir]).add;
+
+
+    ~dirt.addModule('compressor', { |dirtEvent|
+        var cRatio = ~cRatio ?? 4;
```
telephon: `var cRatio = ~cRatio ?? { 4 };` or `var cRatio = ~cRatio ? 4`

telephon: both fine, I'd do the latter.

library/default-effects-extra.scd
```diff
@@ -412,5 +412,70 @@ DEFAULT EFFECTS EXTRA
...
+    	var cKnee = 0;
+
+    	signal = In.ar(out, ~dirt.numChannels);
+
+    	signalAmplitude = Amplitude.ar(signal, 0.001, 1.0);
+
+    	in = signal * cTrim.dbamp;
+    	e = in.mean;
+        e = HPF.ar(e, cHpf);
+        e = EnvDetect.ar(e, cAttack, cRelease);
```
telephon: `EnvDetect` requires sc3plugins. We don't assume everybody has them. Maybe you can replace this by `Amplitude`? It will sound different, but maybe also good?

telephon (commented):
Thank you this looks great! There is only one (compatibility) issue we'd need to find a solution for.

library/default-effects-extra.scd
thgrund: Ah, good point! :-) I ported the SuperDirtMixer compressor from here: https://github.com/alikthename/Musical-Design-in-Supercollider/blob/master/16_compression.sc#L45

You see, that `bias` is a control rate. Then this `K2A.ar(...)` is needed, to do the softlip calculcation correctly. In our case `in` is already an audio rate. So you are totally right. It's not needed here and we can apply the softclip directly to `in`.

Will change it.

library/default-effects-extra.scd
thgrund: I will have a look and try it out.

library/default-effects-extra.scd
thgrund: The last one is more clear for me as well. Will change it.

library/default-effects-extra.scd
telephon: Very good. Could you please also add a reference to the source where you ported it from? A one line comment is enough.

library/default-effects-extra.scd
```diff
@@ -412,5 +412,70 @@ DEFAULT EFFECTS EXTRA
...
+	    )
+    }, { ~cThresh.notNil }); // play synth only if at least one of the two was given
+
+
+    SynthDef("compressor" ++ ~dirt.numChannels, { |out, cThresh, cRatio=4 |
+
+    	var local, signal, signalAmplitude, in, t, o, r, gainReduction, e, cLookaheadDelay, compressedSound, makeupGain;
+    	var cKneelag;
+
+        var cAttack = 0.01;
```
telephon: you could expose these parameters as controls, then they could be tweaked.

thgrund: Yes, I am aware of this. I can add this in a separate commit in this pull request. I mentioned this topic in the description of this PR: 

 ```
All these parameters could set from the outside, but are static

https://github.com/musikinformatik/SuperDirt/pull/314/files#diff-09ee35f9f1ac58413eab41c2ea003e41c624d845d7fdf1e1b9625836aabd59faR438-R442
```

File: library/default-effects-extra.scd (hunk 3 above)
Comment (thgrund) [reply library/default-effects-extra.scd]: Sounds good to me!

File: library/default-effects-extra.scd (hunk 2 above)
Comment (thgrund) [reply library/default-effects-extra.scd]: Done

File: library/default-effects-extra.scd (hunk 1 above)
Comment (thgrund) [reply library/default-effects-extra.scd]: Done, I changed this line and added some reference materials to the SynthDef

**Comment by thgrund** [comment]
@telephon could you please have a look again and see, whether everything is okay now?

**Comment by telephon** [comment]
@thgrund thank you, thos looks very good. I hope leaving out the `EnvDetect` wasn't changing the thing to the worse? Just now I notice the the whole file has other effects that also depend on sc3-plugins extensions, so it would actually be fine to use it if necessary (sorry!).

**Review by telephon — CHANGES_REQUESTED** [review]
@thgrund There are some whitespace issues still – you can easily fix them by setting the editor to tabs instead of spaces (sorry! :-)), selecting all and hitting tab.

Screenshot 2025-06-22 at 11 31 19

**Comment by thgrund** [comment]
> @thgrund There are some whitespace issues still – you can easily fix them by setting the editor to tabs instead of spaces (sorry! :-)), selecting all and hitting tab.
> Screenshot 2025-06-22 at 11 31 19

Select all + Tab: I learned something new :-D Done

**Comment by thgrund** [comment]
> @thgrund thank you, thos looks very good. I hope leaving out the `EnvDetect` wasn't changing the thing to the worse? Just now I notice the the whole file has other effects that also depend on sc3-plugins extensions, so it would actually be fine to use it if necessary (sorry!).

No problemo :-) TBH I didn't noticed a big difference, but I keep this in mind. Having a sc3-plugins independent compressor is not the worst thing to have :-) 

There is only one open point on my side. Because I added a auto makeup gain, the loudness is not alway optimal (but this is a general topic regarding compression). What would you recommend? Shall I keep the auto makeup gain? Make it optional and overwrite it with an optional parameter or adding the manual makeup gain property? WDYT?

**Review by telephon — APPROVED** [review]

**Comment by telephon** [comment]
> There is only one open point on my side. Because I added a auto makeup gain, the loudness is not alway optimal (but this is a general topic regarding compression). What would you recommend? Shall I keep the auto makeup gain? Make it optional and overwrite it with an optional parameter or adding the manual makeup gain property? WDYT?

Ah good – I'll think about it a little

**Comment by telephon** [comment]
If you wanted to make it switchable, you could add a parameter `makeUpForGain`, which can be set to zero if you don't want to make up for the gain:

```
(cThresh.ampdb * (1 - (1 / cRatio)) * makeUpForGain).neg.dbamp
```

**Comment by thgrund** [comment]
> If you wanted to make it switchable, you could add a parameter `makeUpForGain`, which can be set to zero if you don't want to make up for the gain:
> 
> ```
> (cThresh.ampdb * (1 - (1 / cRatio)) * makeUpForGain).neg.dbamp
> ```

Hey @telephon I had the chance to think about this topic. I mean it is not an issue with the auto makeup gain, when quiet sounds will be quiet. The compressor will reduce the amplitude of the loudest signal (if it's in the threshold range) and then increase the amplitude of the overall signal, to nearly match the original sound. 

So I think it makes no sense to disable the auto makeup gain. But I added a new parameter `cGain`. With this you can increase the amplitude of the result signal.

Original
```SuperCollider
(type: \dirt, dirt: ~dirt, s: \superpiano, note: 0, delta: 2.0).play
```

Quiet signal with compression and cGain:

```SuperCollider
(type: \dirt, dirt: ~dirt, s: \superpiano, note: 0, cThresh: 0.001, gain: 0.4, cGain: 10, delta: 2.0).play
 ```

WDYT?

File: library/default-effects-extra.scd [hunk 5]
```diff
@@ -412,5 +412,74 @@ DEFAULT EFFECTS EXTRA
 			hpfCutoffFreq
 		);
 		ReplaceOut.ar(out, signal)
+    }, [\ir]).add;
```
telephon, line 415: actually, let's shorten it:
```supercollider

~dirt.addModule('compressor', { |dirtEvent|
    dirtEvent.sendSynth('compressor' ++ ~dirt.numChannels,
        [
            cThresh: ~cThresh,
            cRatio: ~cRatio ? 4,
            cAttack: ~cAttack ? 0.01,
            cRelease: ~cRelease ? 0.1,
            cTrim: ~cTrim ? 0,
            cHpf: ~cHpf ? 50,
            cKnee: ~cKnee ? 0,
            cGain: ~cGain ? 1,
            out: ~out
        ]
    )
}, { ~cThresh.notNil }); // play synth only if ~cThresh is given
```

telephon (approved):
I think it is very good like this, thank you! (just this one simplification)

library/default-effects-extra.scd
thgrund: Done

thgrund:
@telephon then we are good to go! :-) It looks like you need to merge this PR? I am not able to do so, even after your approval.

telephon:
Yes, I'll merge. I just approved to signal to you that we are getting there. Thanks for your patience!

